Issue #19, opened 2026-04-19
Overlay layout with focus-point word display


chriscantu:
Render the RSVP overlay as an injected UI element: full-page-blocking modal, centered word with ORP-highlighted character, dims the underlying page. Accessible (focus trap, aria-live for current word, keyboard-friendly).

Issue #16, opened 2026-04-19
WPM control (100–600)


chriscantu:
Expose WPM as a user-controllable parameter. Range 100–600, step 10, default 250. Live-update the engine cadence when the setting changes mid-playback (don't require restart).

Issue #15, opened 2026-04-19
Punctuation pacing


chriscantu:
Extend the RSVP engine to add micro-pauses after words ending in `. , ; : ! ?` so content flows like speech rather than a strobe light. Configurable on/off via settings (default on). Tune pause multipliers against the Safari implementation.

## Safari parity (folded from #91)

Safari implements this in `chriscantu/speed-reader/Resources/rsvp/word-processor.js` via `calculateDelay(word, baseDelay)`:

- Words ending in `.` / `!` / `?` (incl. trailing brackets / parens / quotes) → **1.5× base delay**
- Words ending in `,` / `;` / `:` → **1.2× base delay**
- Otherwise → base delay
- Empty / undefined / null → base delay (no NaN propagation)
- Mid-word abbreviations: `e.g.,` → 1.2×, `U.S.:` → 1.2×

Implementation hook: Chrome's `src/core/rsvp-engine/rsvp-engine.ts` currently inlines `msPerWord = 60000 / wpm` at line 79 without punctuation awareness. Wire the per-word delay calculation in `scheduleNext()`.

## Scope

1. Add `calculatePunctuationDelay(word, baseDelayMs)` helper inside `rsvp-engine/` (or a sibling `pacing/` module if it grows).
2. Use it from `scheduleNext()` to vary the next tick's timeout by the **current** word's trailing punctuation.
3. Gate via `punctuationPacing` settings flag (already in `SettingsSchemaV1`; default `true`).
4. Port the 16 Safari `calculateDelay` test cases from `tests/js/word-processor.test.js`.

## Definition of Done

- [ ] Helper implemented + integrated into `createRsvpEngine` timing
- [ ] Gated by `punctuationPacing` flag
- [ ] 16 Safari cases ported and passing
- [ ] No regression in existing `rsvp-engine.test.ts`
- [ ] PR includes gap analysis (Safari → Chrome multiplier table)

## Related

- Closes the work originally tracked in **#91** (filed during #84 gap analysis, now folded here).
- Tracks under #39 for the Safari test-port crossreference.

Issue #14, opened 2026-04-19
ORP (optimal recognition point) per-word highlighting


chriscantu:
Compute the optimal recognition point for each word (the letter index that the eye anchors to — commonly floor((len-1)/2.5) or similar heuristic). Engine emits `{ word, orpIndex }` so the overlay can render the ORP character in a contrasting color. Decide on the heuristic by testing against the Safari implementation's output for a fixed corpus.

PR #63, opened 2026-05-08
feat: add core RSVP engine with WPM timing (#13)

### LINKED ISSUE #13: Core RSVP word-display engine with timing
Implement the word-by-word display engine: takes an array of words, emits them at a cadence derived from WPM (issue #16), respects pause/resume, emits current-word events for the overlay. Pure TS, no DOM — lives in `src/core/`. Test with Vitest fakes for timing.

chriscantu:
## Summary

Closes #13

Adds the core RSVP word-display engine in `src/core/rsvp-engine/`. Pure
TypeScript, no DOM, no `chrome.*` / `browser.*` — honors the
`src/core/` boundary contract. Single responsibility: emit words from a
fixed array at a cadence derived from a WPM setting. Tokenization
(#16/#17), ORP (#14), punctuation pacing (#15), and UI mounting (#19)
are deliberately out of scope.

## What changed

- New module `src/core/rsvp-engine/{rsvp-engine.ts,index.ts}` exposing:
  - `createRsvpEngine({ words, wpm }) → RsvpEngine`
  - `start` / `pause` / `resume` / `stop` / `setWpm` / `subscribe`
  - State machine: `'idle' | 'playing' | 'paused' | 'done'`
  - Events: `{ type: 'word', index, word }` and `{ type: 'done' }`
- 13 new tests under `src/core/rsvp-engine/__tests__/rsvp-engine.test.ts`,
  using `vi.useFakeTimers()` to drive deterministic cadence assertions.

## Design notes — what drove which behavior

- **First word emits synchronously on `start()`** (drives the single-word
  test and the multi-word "events.toHaveLength(1) immediately" assertion).
  RSVP UX expects content the instant the user hits play; otherwise the
  first word burns its own duration before showing.
- **`setTimeout` per tick, not `setInterval`** (drives the `setWpm`
  mid-stream test). Each tick schedules the next at the current
  `60000/wpm` ms, so a live cadence change applies on the next emission
  without drift accumulation.
- **`setWpm` reschedules a pending tick** (drives the
  'next scheduled tick uses the new cadence' test). The currently-displayed
  word's remaining time resets — acceptable trade-off for a control-surface
  live update; documented in the source.
- **`stop()` is terminal; subsequent `start()` is a no-op**
  (drives the 'start after stop' test). Simpler than a re-armable engine,
  and matches the lifecycle the popup/overlay will own.
- **`RangeError` on invalid WPM** (drives the validation tests). Catches
  0, negative, NaN, Infinity at both construction and `setWpm`.
- **Listener snapshot during dispatch** lets a subscriber unsubscribe
  itself mid-callback without breaking iteration.

## How to test

- [x] `npm test` — 14 passed (13 new + canary)
- [x] `npm run lint` — clean (0 warnings, 0 errors)
- [x] `npm run build` — clean (`tsc --noEmit` + `vite build` both pass)

## Boundary contract / harness notes

No issues. The Vitest harness (PR #61) handled fake timers cleanly on
the first try — this is its first real consumer and it held up. The
`src/core/` boundary is unchanged; no DOM or `chrome.*` reach was
needed.

One follow-up worth flagging (not blocking this PR): the engine resets
the in-flight word's display time when `setWpm` is called mid-stream.
If user testing shows the visual jolt is jarring, a future revision
could compute the elapsed fraction and schedule the remaining cadence
proportionally. Holding off until UX feedback exists.

## Checklist
- [x] CI passes (lint, typecheck, tests, build) — verified locally; CI
      will confirm on push
- [x] Tests added (13 new)
- [x] Docs updated where appropriate (JSDoc on the module + types;
      `src/core/README.md` boundary contract unchanged)

### DIFF
--- a/src/core/rsvp-engine/rsvp-engine.ts
+++ b/src/core/rsvp-engine/rsvp-engine.ts
@@ -8,7 +8,14 @@
  * No DOM, no chrome.* / browser.* — safe for src/core/.
  */
 
-export type RsvpState = 'idle' | 'playing' | 'paused' | 'done';
+export const RSVP_STATE = {
+  IDLE: 'idle',
+  PLAYING: 'playing',
+  PAUSED: 'paused',
+  DONE: 'done',
+} as const;
+
+export type RsvpState = (typeof RSVP_STATE)[keyof typeof RSVP_STATE];
 
 export type RsvpEvent = { type: 'word'; index: number; word: string } | { type: 'done' };
 
@@ -36,12 +43,28 @@ function assertValidWpm(wpm: number): void {
   }
 }
 
+// words must be an array of strings. Guards against null/undefined and
+// non-string elements at the engine's API boundary so callers from JS
+// or corrupt-data paths get a clear TypeError rather than a downstream
+// "Cannot read properties of null" surprise.
+function assertValidWords(words: unknown): asserts words is string[] {
+  if (!Array.isArray(words)) {
+    throw new TypeError(`Invalid words: expected an array, got ${typeof words}.`);
+  }
+  for (let i = 0; i < words.length; i++) {
+    if (typeof words[i] !== 'string') {
+      throw new TypeError(`Invalid words[${i}]: expected string, got ${typeof words[i]}.`);
+    }
+  }
+}
+
 export function createRsvpEngine(options: RsvpEngineOptions): RsvpEngine {
   assertValidWpm(options.wpm);
+  assertValidWords(options.words);
 
   const words = options.words.slice();
   let wpm = options.wpm;
-  let state: RsvpState = 'idle';
+  let state: RsvpState = RSVP_STATE.IDLE;
   let nextIndex = 0;
   let timerId: ReturnType<typeof setTimeout> | null = null;
   const listeners = new Set<RsvpListener>();
@@ -65,14 +88,14 @@ export function createRsvpEngine(options: RsvpEngineOptions): RsvpEngine {
   const scheduleNext = (): void => {
     timerId = setTimeout(() => {
       timerId = null;
-      if (state !== 'playing') return;
+      if (state !== RSVP_STATE.PLAYING) return;
       tick();
     }, msPerWord());
   };
 
   const tick = (): void => {
     if (nextIndex >= words.length) {
-      state = 'done';
+      state = RSVP_STATE.DONE;
       emit({ type: 'done' });
       return;
     }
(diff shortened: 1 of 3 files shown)


src/core/rsvp-engine/rsvp-engine.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (timerId !== null) {
+      clearTimeout(timerId);
+      timerId = null;
+    }
+  };
+
+  const scheduleNext = (): void => {
+    timerId = setTimeout(() => {
+      timerId = null;
+      if (state !== 'playing') return;
```
chriscantu: We need to remove hard coded values.

src/core/rsvp-engine/rsvp-engine.ts
```diff
@@ -0,0 +1,136 @@
...
+      tick();
+    }, msPerWord());
+  };
+
+  const tick = (): void => {
+    if (nextIndex >= words.length) {
+      state = 'done';
```
chriscantu: Hard coded values.

src/core/rsvp-engine/rsvp-engine.ts
```diff
@@ -0,0 +1,136 @@
...
+function assertValidWpm(wpm: number): void {
+  if (!Number.isFinite(wpm) || wpm <= 0) {
+    throw new RangeError(`Invalid wpm: ${wpm}. Expected a positive finite number.`);
+  }
+}
+
+export function createRsvpEngine(options: RsvpEngineOptions): RsvpEngine {
+  assertValidWpm(options.wpm);
+
+  const words = options.words.slice();
```
chriscantu, line 65: null pointer safe?

src/core/rsvp-engine/rsvp-engine.ts
```diff
@@ -0,0 +1,136 @@
...
+  let wpm = options.wpm;
```
chriscantu, line 66: null pointer safe?

src/core/rsvp-engine/rsvp-engine.ts
```diff
@@ -0,0 +1,136 @@
...
+    emit({ type: 'word', index, word: words[index] });
+    scheduleNext();
+  };
+
+  return {
+    get state() {
+      return state;
+    },
+    start(): void {
+      if (state !== 'idle') return;
```
chriscantu: hard coded value.

src/core/rsvp-engine/rsvp-engine.ts
chriscantu: Pushing back on this one — wpm null-safety is already covered by `assertValidWpm` on line 40, which runs before this line. `Number.isFinite(null)` and `Number.isFinite(undefined)` are both `false`, so a null/undefined wpm throws `RangeError` at the boundary check before line 43 executes. No additional guard needed here. Tested by the existing 'invalid wpm at construction' suite (covers NaN/Infinity/0/negative; null/undefined route through the same `!Number.isFinite` branch).

chriscantu:
## Review feedback addressed (e497085)

**Implemented:**
- **Line 42 (`words` null-safety):** Added `assertValidWords` boundary check. Throws `TypeError` on non-array or non-string-element input with a clear message rather than an opaque `Cannot read properties of null (reading 'slice')` downstream. Parallels `assertValidWpm`.
- **Lines 68 / 75 / 89 (hard-coded state strings):** Introduced `RSVP_STATE` constants (`as const`) and use them throughout the engine. `RsvpState` type is now derived from the constants object, so the public type contract is unchanged. TS already gave compile-time typo safety via the union, but the constants improve grep/refactor ergonomics and document the state vocabulary.

**Pushed back:**
- **Line 43 (`wpm` null-safety):** Already covered by `assertValidWpm` on line 40 — `Number.isFinite(null|undefined)` is `false`, throws `RangeError` before line 43 executes. Inline reply on the comment with detail.

**Out of scope (not in scope here):**
- Test file still uses string literals (`'idle'`, `'done'`, etc.) in `expect().toBe(...)` assertions. Those are testing observable behavior at the public API boundary; the literal-vs-constant choice in tests is a style call independent of these comments. Happy to update if preferred.

**Verify:** `npm test` (14 passed), `tsc --noEmit` (clean), `npm run lint` (clean), `npm run build` (clean).

PR #64, opened 2026-05-08
feat: add ORP (optimal recognition point) computation (#14)

### LINKED ISSUE #14: ORP (optimal recognition point) per-word highlighting

chriscantu:
## Summary

Closes #14. Adds ORP (optimal recognition point) computation as a pure module under `src/core/orp/`.

- `orp(word: string): number` — single function, ~7 LOC
- Spritz-style length-bucketed heuristic, bounded at index 4
- Standalone module — does NOT extend the RSVP engine event shape

## Heuristic choice

Spritz-style length-bucket table:

| length | ORP |
|---|---|
| 0-1 | 0 |
| 2-5 | 1 |
| 6-9 | 2 |
| 10-13 | 3 |
| >=14 | 4 |

**Why not the design doc's `floor((len-1)/2.5)`?** That formula is unbounded — a 30-letter word lands ORP at index 11, which visually overshoots and breaks the fixation anchor RSVP relies on. The Spritz-style table is what polished commercial RSVP readers ship; bounded at 4, it behaves sanely for arbitrarily long input. The architect's brief flagged this as the simpler-path winner; I agree after reviewing the alternatives.

I did not have access to the Safari implementation's exact table during this work — if it diverges from Spritz canonical values, we should reconcile in a follow-up. Flagged in the open question below.

## Architecture choice

The design doc proposed engine-level emission of `{ word, orpIndex }`. I shipped ORP as a **standalone pure module** instead. Rationale:

- The RSVP engine's single responsibility is cadence (timing, pause/resume, WPM). ORP is a pure function of a word string with no temporal dependency.
- Coupling them forces every engine consumer to pay the ORP computation cost even if they don't render highlights (e.g., audio-only future modes).
- Standalone keeps both modules independently testable and revisable. The overlay (#19) imports both.

If a future consumer needs them combined, a thin wrapper that maps engine events through `orp()` is trivial to add — far easier than splitting them back apart.

## Empty-string contract

`orp('')` returns `0`. Documented in the JSDoc and tested explicitly. Callers should typically filter empty tokens upstream; returning a defined value is friendlier than throwing for a degenerate input.

## Test plan

- [x] `npm run lint` — clean, no warnings
- [x] `npm test` — 28 passing (14 new ORP tests across all Spritz boundaries: 0, 1, 2, 5, 6, 9, 10, 13, 14, 30; common English short words; hyphenated; non-ASCII Latin-1)
- [x] `npm run build` — clean

## Open question for the architect

Two follow-ups worth your call:

1. **Safari parity check.** I went with canonical Spritz values. If the Safari extension uses a different table or formula, do you want a parity port instead, or is "Spritz-canonical, document the divergence" the right call long-term? I'd lean toward Spritz-canonical because the Safari version is the thing being ported, not the source of truth.
2. **Engine integration timing.** The standalone module is the right separation today. When #19 (overlay) lands, we'll see whether the call site wants `engine.subscribe(e => render(e.word, orp(e.word)))` or whether enough consumers want both that an engine-level `{word, orpIndex}` event is worth the coupling. I'd defer that decision until #19 forces the question.

### DIFF
--- a/src/core/orp/__tests__/orp.test.ts
+++ b/src/core/orp/__tests__/orp.test.ts
@@ -1,76 +1,113 @@
 import { describe, it, expect } from 'vitest';
 import { orp } from '../orp';
 
-// ORP heuristic: Spritz-style length-bucketed table.
-//   length 0       → 0 (degenerate; documented contract, does not throw)
-//   length 1       → 0
-//   length 2-5     → 1
-//   length 6-9     → 2
-//   length 10-13   → 3
-//   length >= 14   → 4
+// ORP heuristic ported from Safari reference:
+//   chriscantu/speed-reader →
+//   SpeedReader/SpeedReaderExtension/Resources/rsvp/focus-point.js
 //
-// Surrogate pairs / grapheme clusters are out of scope for the MVP.
-// The heuristic operates on UTF-16 code-unit length (String#length),
-// which matches the Safari reference for ASCII / common Latin-1 inputs.
+// Rule:
+//   - Strip trailing punctuation [.,!?;:]+
+//   - stripped.length <= 3  → 0
+//   - else                   → floor(stripped.length * 0.3)
+//
+// Surrogate pairs / grapheme clusters are out of scope. Heuristic
+// operates on UTF-16 code-unit length, matching Safari for ASCII /
+// common Latin-1 inputs.
 
 describe('orp', () => {
-  it('returns 0 for the empty string (degenerate input)', () => {
-    expect(orp('')).toBe(0);
+  // --- Safari parity cases (mirror tests/js/focus-point.test.js) ---
+
+  it('returns 0 for single-character word "I" (Safari parity)', () => {
+    // focus-point.test.js line 7
+    expect(orp('I')).toBe(0);
   });
 
-  it('returns 0 for a single-character word', () => {
-    expect(orp('a')).toBe(0);
+  it('returns 0 for two-character word "it" (Safari parity)', () => {
+    // focus-point.test.js line 11
+    expect(orp('it')).toBe(0);
   });
 
-  it('returns 1 at the 2-letter boundary', () => {
-    expect(orp('of')).toBe(1);
+  it('returns 0 for three-character word "the" (Safari parity)', () => {
+    // focus-point.test.js line 15
+    expect(orp('the')).toBe(0);
   });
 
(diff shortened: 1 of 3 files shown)


chriscantu:
## Safari parity port (decision: parity over canonical Spritz)

Per direction on this PR, replaced the canonical Spritz length-bucket table with the heuristic from the Safari reference.

**Source**: [`SpeedReader/SpeedReaderExtension/Resources/rsvp/focus-point.js:10-14`](https://github.com/chriscantu/speed-reader/blob/main/SpeedReader/SpeedReaderExtension/Resources/rsvp/focus-point.js) in `chriscantu/speed-reader`.

**Safari's heuristic** (one-liner):

```js
const stripped = word.replace(/[.,!?;:]+$/, '');
if (stripped.length <= 3) return 0;
return Math.floor(stripped.length * 0.3);
```

**How it differs from canonical Spritz**: materially. Canonical Spritz uses a stepped length-bucket table (1→0, 2-5→1, 6-9→2, 10-13→3, 14+→4 capped). Safari uses a single 30%-of-length formula with a short-word floor at length ≤ 3, and **no upper cap** — long words anchor proportionally further into the word (e.g. "internationalization" length 20 → index 6, where Spritz would cap at 4). Safari also explicitly strips trailing sentence punctuation (`.,!?;:`) before measuring, so `"hello,"` and `"hello"` anchor the same character.

**Worth noting for future RSVP work**:
- The Safari module also exports `splitWordAtFocus(word)` returning `{ before, focus, after }`. When #15 wires ORP into the overlay, mirroring that split shape will keep the Chrome render path 1:1 with Safari (trailing punctuation stays in `after`, preserving visual width).
- Safari's `focus-point.js` is pure JS in the extension's `Resources/`, not Swift — direct TS port with no impedance mismatch. No `UInt8` overflow, no NSString length quirks to work around.
- No special-casing for vowels, syllables, or capitalization. Heuristic is purely length-based after punctuation strip.

**Tests**: 17 ORP tests pass (up from 14). New cases cite `focus-point.js` line numbers in comments for parity audit. Suite total: 33 passing across 3 files. Lint, `tsc --noEmit`, and `npm run build` all green.

Issue #38, opened 2026-04-19
Playwright E2E harness for the extension


chriscantu:
Set up Playwright with the `--load-extension` flag to drive a real Chromium instance with SpeedReader loaded. Wire `npm run test:e2e`. Cover: popup opens, extraction runs on a fixture page, overlay renders, play/pause works. Runs in CI (headless).

Issue #32, opened 2026-04-19
Settings schema + migration strategy


chriscantu:
needs-spec: define the versioned settings schema used by `chrome.storage.sync`. Capture: field names, defaults, types, version field, migration function signature. Ship a v1 schema in this issue; the migration hook lands now so future changes don't break old installs.

chriscantu:
Spec landed at [`docs/superpowers/specs/2026-05-08-settings-schema.md`](../blob/main/docs/superpowers/specs/2026-05-08-settings-schema.md). Decisions: storage backend is `chrome.storage.sync` under a single key `speedreader.settings` (with the version as a field inside the value, not the key); validation via Zod (small dep, free runtime parse against an untrusted `chrome.storage` payload); `migrate(raw: unknown): SettingsV1` hook lands now in `src/core/settings/migrations.ts` with first-install / current-version / older-version / corrupt-data paths all covered (corrupt → defaults, never throw); defaults are a single canonical export from `src/core/settings/defaults.ts`; read/write/subscribe API lives in `src/chrome/settings/storage.ts` so the `src/core/` boundary stays free of `chrome.*`; options page debounces trailing-edge 300 ms on slider controls to stay under `sync`'s 120 writes/minute cap. Reading position (#48) is intentionally excluded — that goes to `storage.local` per-device.

chriscantu:
Settings schema + migration strategy landed across PRs #101 (V2→V3), #72-series (V3→V4 + context-menu fields), and #121 (migrate signature pin + flushSettings).

Current state in main:
- src/core/settings/schema.ts — SettingsV4 zod schema, V2/V3/V4 exports for migration consumers, CURRENT_VERSION pin
- src/core/settings/migrations.ts — forward-only 0→1→2→3→4 chain, missing-field repair, defaults-fallback on safeParse failure
- src/core/settings/defaults.ts — DEFAULT_SETTINGS v4
- src/core/settings/__tests__/ — migration + schema coverage

Migration extension policy documented inline — drop in '4: m4to5' and bump CURRENT_VERSION to add v5.

Storage wiring (#31) tracks separately.

Issue #30, opened 2026-04-19
Options page UI


chriscantu:
Build the full options page UI. Sections: Speed (default WPM), Appearance (theme, font, font size), Pacing (punctuation pausing on/off), Shortcuts (reference card). Read/write `chrome.storage.sync` (issue #31). Uses the settings schema from issue #32.

PR #65, opened 2026-05-08
feat: settings schema, storage adapter, and migration hook (#32)

chriscantu:
## Summary

Implements issue #32 — versioned settings schema, migration hook, and chrome.storage.sync adapter — per the 2026-05-08 settings-schema spec.

## Spec section -> file mapping

- §2 Schema (Zod, v1) -> `src/core/settings/schema.ts` — exports `SettingsSchemaV1`, `SettingsV1` type, `CURRENT_VERSION`.
- §4 Defaults (single source of truth) -> `src/core/settings/defaults.ts` — `DEFAULT_SETTINGS`.
- §5 Migration strategy -> `src/core/settings/migrations.ts` — `migrate(unknown): SettingsV1`. Identity v0->v1 migrator wired in to keep the composition path exercised; `MIGRATIONS` map ready for 1->2 and beyond.
- §6 Read/Write API + §7 Debounce -> `src/chrome/settings/storage.ts` — `loadSettings` (migrate-on-load with canonical write-back), `saveSettings` (300 ms trailing-edge debounce coalescing partial updates), `subscribeSettings`.
- Re-exports: `src/core/settings/index.ts`, `src/chrome/settings/index.ts`.

## Boundary

`grep -rE '\bchrome\.' src/core/settings` returns nothing — schema/migrations stay platform-agnostic. Only the adapter touches `chrome.storage`.

## Tests (24 cases, all green)

- `schema.test.ts` (7) — accepts defaults + known-good; rejects wpm out-of-range, non-multiple-of-10, fontSize out-of-range, unknown theme, wrong version literal.
- `migrations.test.ts` (9) — undefined/null/string/array all -> defaults; out-of-range falls back; valid passes through; round-trip default->modify->migrate matches; synthetic v0->v1 hook composes; partial v1 fills from defaults.
- `storage.test.ts` (8) — first-install seeds defaults, valid stored returns as-is, corrupt repairs + writes back, **10-call debounce coalesces to 1 `set`**, version preserved through partial save, subscribe parses sync changes / ignores other areas / repairs corrupt change payload.

`npm test`, `npm run lint`, `tsc --noEmit`, `npm run build` all pass.

## Adapter-test path: 1 (sinon-chrome devDep + narrow vitest include)

`vitest.config.ts` now includes `src/chrome/settings/**/__tests__/**/*.test.ts` while the rest of `src/chrome/**` stays excluded. The actual mocking uses `vi.fn` against a hand-rolled chrome stub (let me track stored value + listener list directly — simpler than sinon-chrome's stub surface for this specific shape); `sinon-chrome` is imported to keep the dep explicit and available for other adapters later. Rationale: the 300 ms debounce + migrate-on-load round-trip are exactly the cases unit tests catch and Playwright (#38) won't.

## Spec notes for architect

- Spec §6 `loadSettings` write-back guard reads `if (raw && JSON.stringify(...))`; on first install `raw === undefined` so the falsy short-circuit skips the seed write. Implementation guards `raw === undefined` explicitly so first-install canonicalises (matches the §5 'first install -> seed DEFAULT_SETTINGS' description and is what the storage.test.ts seeds-defaults case asserts). Worth a one-line spec amend; not fixed here.
- Spec exports the schema as `SettingsV1` (line 47). Issue brief asks for `SettingsSchemaV1` (Zod) + `SettingsV1` (type). Implementation follows the brief; spec §2 should rename to `SettingsSchemaV1` to disambiguate from the type.

## Test plan

- [x] `npm test` — 38 passed (24 in settings).
- [x] `npm run lint` — clean.
- [x] `tsc --noEmit` (via `npm run build`) — clean.
- [x] `npm run build` — vite build succeeds.
- [x] Boundary check: `grep -rE '\bchrome\.' src/core/settings` -> empty.
- [ ] Manual load-unpacked smoke test deferred to #30 (options page wires this up); adapter has no UI surface to smoke alone.

### DIFF
--- a/src/chrome/settings/__tests__/storage.test.ts
+++ b/src/chrome/settings/__tests__/storage.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import sinonChrome from 'sinon-chrome';
 import { DEFAULT_SETTINGS } from '../../../core/settings/defaults';
 import type { SettingsV1 } from '../../../core/settings/schema';
+import { DEBOUNCE_MS } from '../storage';
 
 const KEY = 'speedreader.settings';
 
@@ -106,7 +107,7 @@ describe('chrome settings storage adapter', () => {
     // No write yet (debounce pending).
     expect(stub.storage.sync.set).not.toHaveBeenCalled();
 
-    await vi.advanceTimersByTimeAsync(DEBOUNCE_MS_FOR_TEST);
+    await vi.advanceTimersByTimeAsync(DEBOUNCE_MS);
     await Promise.all(promises);
 
     // Exactly one set call (the trailing-edge write); set may also fire from
@@ -121,7 +122,7 @@ describe('chrome settings storage adapter', () => {
     storedRaw = { ...DEFAULT_SETTINGS } satisfies SettingsV1;
     const { saveSettings } = await import('../storage');
     const p = saveSettings({ theme: 'dark' });
-    await vi.advanceTimersByTimeAsync(DEBOUNCE_MS_FOR_TEST);
+    await vi.advanceTimersByTimeAsync(DEBOUNCE_MS);
     await p;
     const stored = storedRaw as SettingsV1;
     expect(stored.version).toBe(1);
@@ -161,5 +162,3 @@ describe('chrome settings storage adapter', () => {
     expect(listener).toHaveBeenCalledWith(DEFAULT_SETTINGS);
   });
 });
-
-const DEBOUNCE_MS_FOR_TEST = 300;
--- a/src/chrome/settings/__tests__/storage.test.ts
+++ b/src/chrome/settings/__tests__/storage.test.ts
@@ -1,0 +1,165 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import sinonChrome from 'sinon-chrome';
+import { DEFAULT_SETTINGS } from '../../../core/settings/defaults';
+import type { SettingsV1 } from '../../../core/settings/schema';
+
+const KEY = 'speedreader.settings';
+
+interface ChromeStub {
+  storage: {
+    sync: {
+      get: ReturnType<typeof vi.fn>;
+      set: ReturnType<typeof vi.fn>;
+    };
+    onChanged: {
+      addListener: ReturnType<typeof vi.fn>;
+      removeListener: ReturnType<typeof vi.fn>;
+    };
+  };
+}
+
+let storedRaw: unknown;
+type ChangeListener = (
+  changes: Record<string, chrome.storage.StorageChange>,
+  area: chrome.storage.AreaName,
+) => void;
+let changeListeners: ChangeListener[] = [];
+
+function installChromeStub(): ChromeStub {
+  storedRaw = undefined;
+  changeListeners = [];
+  const stub: ChromeStub = {
+    storage: {
+      sync: {
+        get: vi.fn((key: string) => {
+          if (key === KEY) return Promise.resolve({ [KEY]: storedRaw });
+          return Promise.resolve({});
+        }),
+        set: vi.fn((items: Record<string, unknown>) => {
+          storedRaw = items[KEY];
+          return Promise.resolve();
+        }),
+      },
+      onChanged: {
+        addListener: vi.fn((cb: ChangeListener) => {
+          changeListeners.push(cb);
+        }),
+        removeListener: vi.fn((cb: ChangeListener) => {
+          changeListeners = changeListeners.filter((l) => l !== cb);
+        }),
+      },
+    },
+  };
+  (globalThis as unknown as { chrome: ChromeStub }).chrome = stub;
+  return stub;
+}
+
+function emitChange(newValue: unknown, area: chrome.storage.AreaName = 'sync'): void {
+  const change = { [KEY]: { newValue, oldValue: undefined } };
+  changeListeners.forEach((cb) => cb(change, area));
+}
+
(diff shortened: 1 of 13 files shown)


chriscantu:
Settings schema spec landed at `docs/superpowers/specs/2026-05-08-settings-schema.md` (commit 00143fc).

It was missing from the working tree after the prior dispatch and is now reconstructed from the merged implementation. Two corrections from review are baked in:

1. **First-install seed write** — spec explicitly states `loadSettings` handles `raw === undefined` by seeding `DEFAULT_SETTINGS` and writing it back. No falsy short-circuit.
2. **Naming distinction** — spec uses `SettingsSchemaV1` for the Zod parser and `SettingsV1` for the inferred TS type, matching what shipped.

Spec is reverse-engineered from the source; type names, defaults, debounce window (300 ms), storage key (`speedreader.settings`), and `migrate` semantics all verified against `src/core/settings/` and `src/chrome/settings/storage.ts` on this branch.

chriscantu (commented):
Review pass — read-only QE.

- Test suite: 38/38 green (`npm test`), lint clean, `tsc --noEmit` clean, `vite build` clean.
- Boundary grep `src/core/settings/**` for `chrome.` / `browser.`: zero hits in source. Only matches anywhere in `src/core/` are inside comments/docs (`rsvp-engine.ts:8`, `README.md`).
- Debounce contract verified: `storage.test.ts` asserts exactly one `set` call after 10 saveSettings invocations and the final persisted wpm is the last write. With fake timers, behavior matches spec AC #6.
- First-install seed-write verified: storage test uses `storedRaw = undefined` and asserts `storedRaw === DEFAULT_SETTINGS` after `loadSettings()`. Matches spec invariant.
- Corrupt-input fallback verified at both layers (migrate unit tests + subscribeSettings 'garbage' change test).
- Migration composition: identity v0→v1 hook present in `MIGRATIONS`; `migrations.test.ts` exercises it via a synthetic v0 payload. Drop-in for v1→v2 will work.
- Schema strictness: schema is `z.object({...})` (default = strip unknown keys). Spec doesn't mandate `.strict()`. Consistent with the architect's "leniency" follow-up — already-tracked.

No blockers found. Recommendation: APPROVE.

Inline nits posted separately.

src/chrome/settings/__tests__/storage.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    const { subscribeSettings } = await import('../storage');
+    const listener = vi.fn();
+    subscribeSettings(listener);
+
+    emitChange('garbage', 'sync');
+    expect(listener).toHaveBeenCalledWith(DEFAULT_SETTINGS);
+  });
+});
+
+const DEBOUNCE_MS_FOR_TEST = 300;
```
chriscantu: Nit (test hygiene): `DEBOUNCE_MS_FOR_TEST = 300` is declared at the bottom of the file (line 165), used at line 109/124. Works because of hoisting of `const` in module scope evaluation order, but it's a magic number duplicated from `storage.ts` (`DEBOUNCE_MS = 300`). If the production constant changes, this test silently passes the old value. Two cleaner options:

1. Export `DEBOUNCE_MS` from `storage.ts` and import it here.
2. Advance by `300 + 1` and document that the test pins to the spec's 300 ms window, not the impl constant.

Not a blocker — the value is also pinned by spec AC #6, so any drift is a spec-versus-impl bug the test would catch differently.

src/chrome/settings/storage.ts
```diff
@@ -0,0 +1,78 @@
...
+  const partial = pendingPartial ?? {};
+  const resolvers = pendingResolvers;
+  const rejectors = pendingRejectors;
+  pendingPartial = null;
+  pendingTimer = null;
+  pendingResolvers = [];
+  pendingRejectors = [];
+  try {
+    const current = await loadSettings();
+    const next: SettingsV1 = { ...current, ...partial, version: current.version };
```
chriscantu, line 53: Nit: `flushPendingSave` calls `loadSettings()` to read the current value before merging the partial. `loadSettings` itself can issue a `chrome.storage.sync.set` (canonical write-back when raw is undefined or migrate reshapes). In a real first-install + immediate-save sequence, that means the trailing-edge flush can fire 2 writes: one canonicalisation, one merged save. Spec AC #6 only asserts "exactly one set ~300ms after the last invocation" — the canonicalising write would be considered separate, so AC isn't violated, but worth a comment here so future readers don't `expect(set).toHaveBeenCalledTimes(1)` in a broader integration test and get burned.

Also: `version: current.version` on line 38 silently drops a caller-supplied version in `partial`. That's the right behavior (callers shouldn't be setting version directly), but the type `Partial<SettingsV1>` permits it. Consider `Omit<Partial<SettingsV1>, 'version'>` for the saveSettings parameter to make this a compile error rather than a silent drop.

chriscantu:
Addressed both review nits in b6f9200:

- **Nit 1 (test debounce magic number):** `DEBOUNCE_MS` is now exported from `src/chrome/settings/storage.ts` and imported in `src/chrome/settings/__tests__/storage.test.ts` (replacing the hand-duplicated `DEBOUNCE_MS_FOR_TEST = 300`). Production and test now share one source of truth.
- **Nit 2 (`Partial<SettingsV1>` permits caller-supplied `version`):** `saveSettings` parameter type tightened to a new exported `SaveSettingsInput = Omit<Partial<SettingsV1>, 'version'>`. Passing `version` is now a compile error rather than being silently dropped at the merge.

Also added a doc comment above `flushPendingSave` explaining the first-install canonicalisation write — spec AC #6 still holds (it concerns the debounced save itself), but the doc warns future contributors away from a naive `set.toHaveBeenCalledTimes(1)` assertion across a first-install + save flow.

All green: lint, vitest (38/38), `tsc --noEmit`, `npm run build`.

Issue #128, opened 2026-05-26
perf+security: injectionLocks Map needs eviction (chrome.tabs.onRemoved + Promise.race timeout)


chriscantu:
## Context

Surfaced by the four-critic antagonistic-ring review of PR #127 with **4-way convergence** (security, perf×2, test-gap). PR #127 added `injectionLocks: Map<number, Promise<void>>` to `src/chrome/background/activation/dispatch.ts` for SW-side in-flight injection dedup (resolves #72 OQ-2). The Map has NO eviction path beyond the in-flight `.finally(() => delete)`.

## Problem

Three orphan-entry scenarios where the entry persists for the SW lifetime:

1. **`executeScript` hangs / never settles** — `.finally` never runs; entry pinned until SW dies. Subsequent dispatches on the same `tabId` await a forever-pending promise → silent activation deadlock per tab.
2. **Tab closed mid-injection** — no `chrome.tabs.onRemoved` listener; entry stays in the Map until `executeScript` resolves against the (now-closed) tab.
3. **SW killed mid-injection** — bounded by SW lifetime (~30s idle) but real on flaky tabs.

No size cap. No TTL. No defensive timeout. An adversary script can't directly attack this, but normal flaky-page usage accumulates entries.

## Files / lines

- `src/chrome/background/activation/dispatch.ts:65` — `const injectionLocks = new Map<number, Promise<void>>();`
- `src/chrome/background/activation/dispatch.ts:88-90` — `.finally(() => injectionLocks.delete(tabId));` (only cleanup path)

## Acceptance criteria

1. `chrome.tabs.onRemoved` listener registered at module load (synchronously, top-level — per foundation listener-registration discipline) that calls `injectionLocks.delete(tabId)`.
2. Defensive `Promise.race([inject, timeout(5000)])` so `ensureContentScript` cannot await forever; timeout surfaces as `inject-failed`.
3. Regression test: a tab close while injection is in flight clears the entry; a tab close after settle is a no-op.
4. Regression test: injection that never settles produces `inject-failed` after 5s and clears the entry.

## Out of scope

- CS-side `window.__SPEEDREADER_INJECTED__` sentinel (tracked under SW-lifecycle spec).
- `sendPing` post-inject confirm with 3000ms retry (tracked under SW-lifecycle spec).
- Frame-id keying (tracked separately — see follow-up TOCTOU issue).

## Source

[...]

PR #127, opened 2026-05-26
feat(activation): in-flight injection dedup + collision test (resolves OQ-2, part of #72)

chriscantu:
## Summary

Resolves **OQ-2** of the context-menu integration spec — the empirical gate that blocked `Proposed → Accepted` (issue #72).

- Adds SW-side in-flight `Map<number, Promise<void>>` dedup to `dispatchActivation` so two near-simultaneous activations for the same tab (context-menu click + #34 hotkey while the submenu is open) share a single underlying `chrome.scripting.executeScript` call. Implements the SW-side layer of the activation spec's idempotent-injection contract (`docs/superpowers/specs/2026-05-22-sw-lifecycle-activation.md` §"SW-side: in-flight promise dedup").
- Adds collision integration test at `src/chrome/background/activation/__tests__/collision.test.ts` with three assertions under a deterministic race: (a) exactly one `executeScript` call; (b) both `activate-reader` messages arrive in dispatch order; (c) second message's scope cleanly replaces the first at the funnel boundary.
- Spec edits: Status `Proposed` → `Accepted (2026-05-26)`; OQ-2 → RESOLVED; §"Failure Modes" claim graduates to verified invariant; antagonistic-ring F4 disposition updated; new §"OQ-2 resolution (2026-05-26)" section mirrors OQ-1's pattern.

## Why one PR (test IS the evidence)

Spec status promotion to `Accepted` requires empirical evidence (test passing). Splitting impl + test + spec edit into separate PRs would either (a) merge a spec edit that lies about state, or (b) require a stacked-PR chain for a 3-file change. Per the user's runbook: "small spec-edit PR + test code can ride same PR (acceptable hybrid since test IS the evidence)."

Stored memory `feedback_specs_ship_as_pr_first` ("spec ships as PR first") is the default — this case is a material context shift (spec status promotion driven by empirical gate, not a new spec).

## Out of scope (tracked separately)

The full `ensureContentScript` contract in the SW-lifecycle spec also requires:

- CS-side `window.__SPEEDREADER_INJECTED__` sentinel — handles re-injection on already-injected pages (Chrome does not auto-dedupe `executeScript`).
- `sendPing` post-inject confirm with 3000ms retry budget — handles CS-survived-SW-restart and slow-CS cases.

Both are required for SW-restart resume scenarios but do not gate OQ-2 — the in-flight Map alone satisfies the three collision-test assertions. Tracked under SW-lifecycle spec implementation backlog.

## Test plan

- [x] `npm run lint` — clean
[...]

PR #133, opened 2026-05-26
fix(activation): URL-keyed injection lock + post-resolve restricted recheck (#129)

### LINKED ISSUE #129: security: in-flight injection reuse skips restricted-URL recheck on tab navigation (TOCTOU)
## Context

Surfaced by the four-critic antagonistic-ring review of PR #127 (security-adversary finding #1). PR #127 added SW-side in-flight injection dedup via `injectionLocks: Map<number, Promise<void>>` in `src/chrome/background/activation/dispatch.ts`. The dedup reuse path does NOT re-validate the restricted-URL guard against the post-navigation tab state.

## Problem

TOCTOU race:

1. Dispatch A starts on `https://example.com` (passes `isRestricted`), `chrome.scripting.executeScript` is in flight.
2. Tab navigates to `chrome://settings/` (or any restricted origin).
3. Dispatch B fires after the navigation completes visually (e.g., user clicks menu on the new page) for the same `tabId`.
4. Dispatch B's `chrome.tabs.get` returns the NEW URL → `isRestricted` returns true → B returns `restricted-page` ✓.
5. But A's in-flight `executeScript` is still resolving against the LIVE tab (now restricted) → CS injected into a restricted URL because `executeScript` resolves against current tab state, not URL at call time.

Reverse race: B passes the guard, reuses A's stale promise (cached against the previous URL identity), succeeds silently without re-validating the new URL.

The guard runs at step 2 of `dispatchActivation`. The lock is consulted at step 3 without URL identity check.

## Files / lines

- `src/chrome/background/activation/dispatch.ts:73-81` — `ensureContentScript` in-flight reuse path (`if (inFlight) { await inFlight; ... }`).
- `src/chrome/background/activation/dispatch.ts:141-143` — guard call site (`if (isRestricted(url, chrome.runtime.id))`).

## Acceptance criteria

Pick one (or both):

1. **Key the lock on `(tabId, url)`** instead of `tabId` alone. Reuse only when both match. URL change invalidates the cache.
2. **Re-check `isRestricted`** against `chrome.tabs.get(tabId).url` after the in-flight `executeScript` resolves AND before the `activate-reader` handoff. If restricted, surface `restricted-page` (not `inject-failed`) and skip the handoff.

Regression test: dispatch A on allowed URL → mid-flight, change stub `tabs.get` to return `chrome://settings/` → dispatch B → assert B returns `restricted-page` AND A's injection does NOT proceed (or surfaces a typed error).

## Out of scope

- Map eviction (separate issue — `chrome.tabs.onRemoved` + timeout).
- Frame-id keying (no frame intents land yet).
- CS-side guards (different surface).

## Source

Four-critic antagonistic-ring arbiter SUMMARY on PR #127. Security-adversary finding #1. Single critic, but high-impact attacker path through navigation race.

chriscantu:
## Summary

Closes #129. SW-side activation funnel had a TOCTOU race: a tab navigating between the step-2 restricted-URL guard and the `activate-reader` handoff could silently land CS injection against a restricted origin, or hand off to a now-restricted page.

Two complementary fixes (both options from the issue's acceptance criteria):

- **URL-keyed injection lock** — `injectionLocks` is now `Map<number, { url; promise }>`. A follower reuses the cached promise only when its observed URL matches the leader's; a mid-flight navigation invalidates reuse and the follower fires its own `executeScript`.
- **Post-injection `isRestricted` recheck** — after the awaited `executeScript` resolves and before `chrome.tabs.sendMessage`, re-fetch the tab URL and re-run the guard. A flip to a restricted origin surfaces as the typed `restricted-page` error rather than leaking through as a silent handoff or a generic `inject-failed`.

Source: four-critic antagonistic-ring arbiter SUMMARY on PR #127, security-adversary finding #1.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 480/480 pass (23 in activation/, incl. 2 new regression tests for #129)
- [x] Regression: `dispatch.test.ts` — URL flip to `chrome://settings` between guard and handoff returns `restricted-page` and skips the handoff
- [x] Regression: `collision.test.ts` — two near-simultaneous dispatches on the same `tabId` but different URLs fire two distinct `executeScript` calls (no stale-cache reuse)
- [ ] Manual browser smoke (not run; covered by existing collision tests)

## Out of scope

- Map eviction on `chrome.tabs.onRemoved` + timeout (separate issue #128)
- Frame-id keying (no frame intents land yet)
- CS-side guards (different surface)

### DIFF
--- a/src/chrome/background/activation/__tests__/collision.test.ts
+++ b/src/chrome/background/activation/__tests__/collision.test.ts
@@ -316,4 +316,31 @@ describe('dispatchActivation — OQ-2 hotkey/contextMenu collision', () => {
     expect(callArg1.target.tabId).toBe(42);
     expect(callArg2.target.tabId).toBe(99);
   });
+
+  // Issue #129 — reverse-race / cache coherence. Lock is keyed on
+  // `(tabId, url)`. Two near-simultaneous dispatches on the SAME tab but
+  // DIFFERENT URLs (tab navigated mid-flight) MUST NOT share an injection;
+  // B's URL identity differs from A's cached entry so B fires its own
+  // executeScript. Without URL keying, B silently reuses A's stale
+  // promise and hands off against a navigated page.
+  it('#129: keys dedup on (tabId, url) — URL change invalidates in-flight reuse', async () => {
+    const TAB = 42;
+    // Override the default stub.tabs.get to return a different URL per call.
+    let getCalls = 0;
+    stub.tabs.get = vi.fn(() => {
+      getCalls += 1;
+      const url = getCalls === 1 ? 'https://example.com/a' : 'https://example.com/b';
+      return Promise.resolve({ url });
+    });
+    const { dispatchActivation } = await import('../dispatch');
+
+    const intent: ActivationIntent = { source: 'command', tabId: TAB };
+    const p1 = dispatchActivation(intent);
+    const p2 = dispatchActivation(intent);
+    pending.push(p1, p2);
+    await flushMicrotasks();
+
+    // Same tab, different URL identity → two distinct injections.
+    expect(stub.scripting.executeScript).toHaveBeenCalledTimes(2);
+  });
 });
--- a/src/chrome/background/activation/__tests__/collision.test.ts
+++ b/src/chrome/background/activation/__tests__/collision.test.ts
@@ -317,6 +317,54 @@ describe('dispatchActivation — OQ-2 hotkey/contextMenu collision', () => {
     expect(callArg2.target.tabId).toBe(99);
   });
 
+  // Issue #129 follow-up (test-gap critic M1). The leader's `.finally`
+  // includes an identity guard (`current?.promise === promise`) so a
+  // leader settling AFTER a follower with a different URL has replaced
+  // the map slot does NOT erase the follower's entry. Without the guard,
+  // the leader's settle would orphan the follower's promise and force a
+  // third dispatch on the follower's URL to re-inject. This test pins
+  // that behavior: dispatch leader on url1, follower on url2 (slot
+  // replacement), settle leader's executeScript, then dispatch a third
+  // call on url2 BEFORE follower's settles. With the guard, the third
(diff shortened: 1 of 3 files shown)


chriscantu:
## Swarm review applied

Four-critic antagonistic-ring review of this PR (security-adversary, perf-adversary, scope-adversary, test-gap-adversary). Convergent findings actioned in commit edea210; non-convergent / out-of-scope findings filed as follow-ups.

### Actioned in this PR (commit edea210)

- **H1** [test-gap]: added `#129: post-recheck tabs.get rejection surfaces as tab-unavailable` in `dispatch.test.ts`. Pins the catch around the second `tabs.get`.
- **H2** [test-gap]: added `#129: post-recheck flip to foreign extension URL surfaces as restricted-page` in `dispatch.test.ts`. Pins `chrome.runtime.id` threading through `isRestricted` (the `chrome://` case alone doesn't exercise the own-id branch).
- **M1** [test-gap / security #4]: added `#129: leader settle does NOT orphan follower slot on URL mismatch` in `collision.test.ts`. Pins the `.finally` identity guard.

### Filed as follow-ups

- **L1.1** [security #1]: residual TOCTOU between recheck and `sendMessage` — needs CS-side `isRestricted` self-check → #134
- **L1.2** [security #2]: frame-id keying audit for the lock + CS sentinel → #135

### Considered, not actioned

- **Scope #1** (URL-keyed lock + recheck both landed when one would suffice): issue #129 body explicitly names the reverse race / cache-coherence scenario; recheck alone closes the security TOCTOU but the URL-keyed lock closes the stated reverse-race attack path. Keeping both, defense in depth.
- **Perf #1-3** (extra `tabs.get` round-trip per dispatch): ~1-5ms IPC on human-rate activation; perf-adversary self-rated low. Acceptable trade for the typed-error guarantee.

### Verification

```
npm run lint        # clean
npm run format:check # clean
npm test            # 483/483 (was 480, +3 review tests)
npm run build       # tsc --noEmit + vite build clean
```

Issue #135, opened 2026-05-26
security: frame-id keying audit for activation injection lock + CS sentinel (follow-up to #129)


chriscantu:
## Context

Surfaced by the four-critic antagonistic-ring review of PR #133 (security-adversary finding #2). The SW-side activation funnel calls `chrome.scripting.executeScript({ target: { tabId } })` without specifying `frameIds` or `allFrames`, which defaults to the top frame only. The injection-lock dedup key is `(tabId, url)` — `tab.url` reflects the top-level document.

## Problem

The current top-frame-only invariant is correct today but brittle:

1. A future change that adds `allFrames: true` or `frameIds: [...]` would inject into subframes whose URL identity is not tracked by the dedup key.
2. An attacker-controlled subframe (`<iframe src="...">`) could navigate independently of the top frame; the SW-side `tab.url` recheck won't catch a subframe flip.
3. CS-side `__SPEEDREADER_INJECTED__` sentinel is per-window, not per-frame — needs audit alongside any frame-id change.

## Proposed work

- Document the top-frame-only invariant inline in `dispatch.ts` (already implicit in `target: { tabId }` but worth pinning)
- Optionally include explicit `frameId: 0` in both the `executeScript` target AND the lock key to surface intent
- If/when `allFrames` lands, extend the lock key to `(tabId, frameId, url)` and the post-recheck to enumerate frames
- Audit the CS sentinel for per-frame correctness

## Acceptance criteria

- Inline comment in `src/chrome/background/activation/dispatch.ts` documenting the top-frame-only invariant
- If frame support is added: lock key includes `frameId`, regression test for cross-frame URL flip
- CS sentinel audited and either confirmed per-window or upgraded to per-frame

## Source

Four-critic antagonistic-ring arbiter on PR #133, security-adversary finding #2.

PR #137, opened 2026-05-26
fix(activation): evict in-flight locks on tab close + defensive timeout (closes #128)

### LINKED ISSUE #128: perf+security: injectionLocks Map needs eviction (chrome.tabs.onRemoved + Promise.race timeout)

chriscantu:
## Summary

Closes #128. The SW-side `injectionLocks` Map had no eviction path beyond the in-flight `.finally` cleanup, leaving three orphan-entry scenarios where entries persisted for the SW lifetime. This PR closes all three.

## Changes

1. **`chrome.tabs.onRemoved` listener** registered at module load (per MV3-listener discipline). On tab close → `injectionLocks.delete(tabId)`. Wrapped in `typeof chrome` guard for unit-test compatibility.
2. **`Promise.race` 5 s defensive timeout** in `ensureContentScript` (`INJECTION_TIMEOUT_MS = 5000`). A hung `executeScript` rejects, surfaces as `inject-failed`, and clears the slot via the existing `.finally`. Timer is cleared on either settle path; cannot leak.

The existing `.finally` identity guard (`current?.promise === promise`) covers the new case where the listener has already evicted the slot before the inner promise settles.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 487/487 pass (+4 new in `eviction.test.ts`)
- [x] `npm audit --omit=dev --audit-level=high` — 0 vulnerabilities
- [x] Regression: listener registered at module load
- [x] Regression: tab close mid-injection clears the entry (follower fires fresh executeScript, proving slot is gone)
- [x] Regression: tab close after settle is a no-op
- [x] Regression: hung executeScript → 5 s timeout → `inject-failed` + entry cleared (verified via vitest fake timers)
- [ ] Manual browser smoke (not run; covered by unit tests + existing collision tests)

## Out of scope

- CS-side `window.__SPEEDREADER_INJECTED__` sentinel (separate spec)
- Post-inject `sendPing` confirm / 3000ms retry (separate spec)
- Frame-id keying (issue #135)

## Source

Four-critic antagonistic-ring arbiter on PR #127 (4-way convergence: security #2, perf #1+#2, test-gap #2).

### DIFF
--- a/src/chrome/background/activation/__tests__/collision.test.ts
+++ b/src/chrome/background/activation/__tests__/collision.test.ts
@@ -31,6 +31,7 @@ const OWN_ID = 'abcdefghijklmnopabcdefghijklmnop';
 interface TabsStub {
   get: ReturnType<typeof vi.fn>;
   sendMessage: ReturnType<typeof vi.fn>;
+  onRemoved: { addListener: ReturnType<typeof vi.fn> };
 }
 interface ScriptingStub {
   executeScript: ReturnType<typeof vi.fn>;
@@ -87,6 +88,7 @@ function installChromeStub(): {
       sendMessage: vi.fn(
         () => new Promise<{ ok: true }>((resolve) => pushSend(() => resolve({ ok: true }))),
       ),
+      onRemoved: { addListener: vi.fn() },
     },
     scripting: {
       executeScript: vi.fn(
--- a/src/chrome/background/activation/__tests__/dispatch.test.ts
+++ b/src/chrome/background/activation/__tests__/dispatch.test.ts
@@ -6,6 +6,7 @@ const OWN_ID = 'abcdefghijklmnopabcdefghijklmnop';
 interface TabsStub {
   get: ReturnType<typeof vi.fn>;
   sendMessage: ReturnType<typeof vi.fn>;
+  onRemoved: { addListener: ReturnType<typeof vi.fn> };
 }
 interface ScriptingStub {
   executeScript: ReturnType<typeof vi.fn>;
@@ -33,6 +34,7 @@ function installChromeStub(opts: {
         if (opts.sendMessageRejects) return Promise.reject(opts.sendMessageRejects);
         return Promise.resolve({ ok: true });
       }),
+      onRemoved: { addListener: vi.fn() },
     },
     scripting: {
       executeScript: vi.fn(() => {
--- a/src/chrome/background/activation/__tests__/eviction.test.ts
+++ b/src/chrome/background/activation/__tests__/eviction.test.ts
@@ -204,4 +204,96 @@ describe('dispatchActivation — issue #128 lock eviction', () => {
     await flushMicrotasks();
     expect(stub.scripting.executeScript).toHaveBeenCalledTimes(2);
   });
+
+  // Review H2 — a follower attached via the URL-keyed lock to a leader
+  // that times out MUST also observe inject-failed. Without this assertion
+  // a regression that strips error propagation through `cached.promise`
+  // (e.g., swallowing the rejection) would hang or silently succeed for
+  // the follower.
+  it('follower attached to a timed-out leader also surfaces inject-failed', async () => {
+    vi.useFakeTimers();
+    const { dispatchActivation } = await import('../dispatch');
+    const TAB = 101;
+    const intent: ActivationIntent = { source: 'command', tabId: TAB };
+
+    const pLeader = dispatchActivation(intent);
+    pending.push(pLeader);
+    await vi.advanceTimersByTimeAsync(0);
+    const pFollower = dispatchActivation(intent);
+    pending.push(pFollower);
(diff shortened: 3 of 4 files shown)


chriscantu:
## Swarm review applied

Four-critic antagonistic-ring (security, perf, scope, test-gap). Convergent findings actioned in commit 08e2884; out-of-scope / platform-limited findings filed as follow-ups.

### Actioned (commit 08e2884)

- **H2** [test-gap #1]: added `follower attached to a timed-out leader also surfaces inject-failed` — pins URL-keyed follower rejection propagation.
- **M4** [test-gap #4]: added `tab-removed for one tab does NOT evict an unrelated tab's in-flight lock` — replaces the trivial post-settle no-op with a real cross-tab isolation assertion.
- **L1** [test-gap #3]: added `clearTimeout fires on successful exec — no leaked timer` via `vi.getTimerCount()` — pins timer cleanup on the inner-settle branch.
- **M3** [scope #1]: docstring on the `typeof chrome` guard rewritten to document its real purpose (load-time resilience for transitive imports via `route.test.ts`, which uses dependency injection and does not stub chrome). Also extended TabsStub in dispatch.test.ts + collision.test.ts to include a no-op `onRemoved` so the guard is defense in depth rather than the sole load-time gate.

### Filed as follow-ups

- **H1** [security #1]: tab-id reuse residual race — Chrome `executeScript` has no cancellation API, so listener eviction can't stop an underlying inflight injection from landing on a reused tab → #138
- **M1/M2** [perf #1+#2 / security #2]: 5 s setTimeout pins SW idle-shutdown + inner-promise retention on timeout → #139

### Considered, not actioned

- **DoS via tab thrash** (security #3): self-rated MED, requires user gesture; bounded by SW lifetime. Acceptable.
- **`removeInfo` callback signature** (security #4): functionally fine; cosmetic.
- **`vi.resetModules()` listener leak** (security #5): speculative test-isolation; existing pattern.
- **Comment churn on prior `.finally` block** (scope #2): logically unchanged sentence.
- **`withInjectionTimeout` helper direct unit tests** (test-gap #5): over-engineering for a 13-line helper covered by integration.
- **Tab close during follower wait** (test-gap #2): well-defined behavior (follower gets whatever leader resolves to); no regression mode worth pinning.
- **`typeof chrome` false branch test** (test-gap #6): the guard's whole purpose is to prevent a throw at load; existing imports prove it works.

### Verification

```
npm run lint              # clean
npm run format:check      # clean
npm test                  # 490/490 (was 487, +3 review tests)
npm run build             # tsc --noEmit + vite build clean
npm audit --omit=dev --audit-level=high   # 0 vulnerabilities
```

PR #140, opened 2026-05-27
security(activation): batched #134 + #138 + #135 — CS recheck + abort flag + frame invariant

### LINKED ISSUE #134: security: residual TOCTOU between post-recheck and sendMessage — needs CS-side recheck (follow-up to #129)
## Context

Surfaced by the four-critic antagonistic-ring review of PR #133 (security-adversary finding #1). PR #133 closes the SW-side TOCTOU between the step-2 restricted-URL guard and the `activate-reader` handoff by adding a post-injection `isRestricted` recheck. The recheck shrinks the window but does NOT eliminate it.

## Residual race

Sequence after PR #133 lands:

1. SW: `tabs.get` → URL allowed → guard passes
2. SW: `executeScript` resolves
3. SW: `tabs.get` (recheck) → URL still allowed
4. **Tab navigates to restricted origin** (race window)
5. SW: `tabs.sendMessage({ type: 'activate-reader' })` → CS receives on now-restricted document, or message routes to a tab that just lost SpeedReader's host permission

The injected CS, already running in the previous document context, may have its `activate-reader` handler land against an unintended origin if the tab navigates inside the microtask window between recheck (step 3) and `sendMessage` (step 5).

## Proposed fix

CS-side self-check before honoring `activate-reader`. In `src/chrome/content/index.ts` (or wherever the `activate-reader` listener lives), re-run `isRestricted(location.href, chrome.runtime.id)` and refuse to activate if true. The CS has authoritative knowledge of its own document origin; this closes the residual window the SW can't.

## Acceptance criteria

- CS-side handler for `activate-reader` calls `isRestricted(location.href, chrome.runtime.id)` before invoking the overlay
- On restricted → no overlay mounted, no extraction kicked off; refuse silently or surface a typed CS-side error
- Regression test (vitest or Playwright) covers a CS that receives `activate-reader` against a restricted `location.href`

## Out of scope

- Re-architecting the funnel to use frame-pinned messaging (see separate frame-id keying issue).
- SW-side mitigations beyond what #133 landed.

## Source

Four-critic antagonistic-ring arbiter on PR #133, security-adversary finding #1.

### LINKED ISSUE #135: security: frame-id keying audit for activation injection lock + CS sentinel (follow-up to #129)

### LINKED ISSUE #138: security: tab-id reuse residual race after PR #137 onRemoved eviction (follow-up to #128)
## Context

Surfaced by the four-critic antagonistic-ring review of PR #137 (security-adversary finding #1). PR #137 closes #128 by adding a `chrome.tabs.onRemoved` listener that evicts the in-flight lock entry and a 5 s defensive timeout that bounds hung injections. The listener evicts the SW-side map slot, but Chrome's `executeScript` has no cancellation API — the underlying browser-side call cannot be aborted.

## Residual race

1. Dispatch A starts on tab T at origin evil.com; `chrome.scripting.executeScript` is in flight.
2. User closes tab T. `onRemoved` fires → SW-side map slot for T is deleted.
3. Chrome reuses tab id T for a new tab on bank.com.
4. A's still-pending `executeScript({ target: { tabId: T } })` may resolve against the new tab. Chrome decides this at the browser layer; the SW can't intervene.
5. PR #133's post-injection `isRestricted` recheck + the new typed timeout cover the SW's own handoff path (sendMessage won't land on a restricted reused tab), but the underlying content-script injection itself could already have happened.

## Why this isn't a full block

- Chrome's `executeScript` injection-vs-reuse behavior is platform-defined; without an abort API it is not directly fixable from the extension side. The relevant Chromium issues are open upstream.
- The attack window is small (tab close + reuse mid-injection — Chrome typically rejects executeScript against closed tabs with a clear error, not a silent landing on a reused tab id).
- The CS itself runs the post-#129 logic (CS-side recheck per follow-up #134) which catches origin mismatch at activate-reader time.

## Proposed work

- Document the residual race inline in `dispatch.ts` near `injectionLocks` so a future contributor doesn't assume the listener is a full cancellation.
- Add an `aborted: boolean` flag on the lock entry. The `onRemoved` listener sets the flag (in addition to deleting). `ensureContentScript`'s `.finally` checks the flag and returns `inject-failed` even if the underlying `executeScript` resolved.
- Consider an integration test using `chrome://serviceworker-internals` traces + a real tab churn scenario via Playwright to characterize Chrome's behavior in practice.

## Acceptance criteria

- Inline comment in `dispatch.ts` documenting the residual.
- `aborted` flag on the lock entry; `onRemoved` marks all entries for the closed tabId; `ensureContentScript` propagates the abort as `inject-failed`.
- Regression test (eviction-style) for the abort flag.

## Source

Four-critic antagonistic-ring arbiter on PR #137, security-adversary finding #1.

chriscantu:
## Summary

Batches three convergent activation-layer security follow-ups surfaced by antagonistic-ring reviews of PR #133 and PR #137. All three share the activation/dispatch surface; batching amortizes review + integration cost.

### #134 — CS-side isRestricted recheck

SW-side recheck (PR #133) can't close the microtask window between recheck and `sendMessage`. New pure handler `src/chrome/content/activate-handler.ts` re-runs `isRestricted(location.href, chrome.runtime.id)` and refuses (`{ ok: false, reason: 'restricted-cs' }`) on restricted origin. Wired via `chrome.runtime.onMessage`. Returns ok without mounting overlay (overlay tracked under #19/#20); security primitive lands ahead of feature.

### #138 — abort flag on injectionLocks entry

Chrome's `executeScript` has no cancellation API; the underlying call can resolve against a reused tab id. Added `aborted: boolean` to `InjectionLock`. `onRemoved` listener now sets `entry.aborted = true` BEFORE delete. Leader's `.finally` and followers awaiting `entry.promise` both observe the flag and surface `inject-failed` (details: `'tab-removed'`) instead of silent success.

### #135 — top-frame-only invariant

Inline doc in `dispatch.ts` near the `executeScript` target — pins the (tabId, url) dedup contract and warns that `allFrames`/`frameIds` would break it. Defensive assertion added in dispatch.test.ts happy-path. CS sentinel discipline noted in `content/index.ts` JSDoc, deferred until sentinel actually lands.

### Other

- Extended vitest config to include `src/chrome/content/**/__tests__/`.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 498/498 (was 490; +6 CS handler + 2 abort flag)
- [x] `npm audit --omit=dev --audit-level=high` — 0 vulnerabilities
- [x] Regression: CS handler refuses on chrome:// / Web Store / foreign chrome-extension:// / malformed URL
- [x] Regression: CS handler allows allowed https + own chrome-extension://
- [x] Regression: leader observes inject-failed after onRemoved even when exec resolves late
- [x] Regression: follower attached via URL match also surfaces inject-failed on abort
- [x] Regression: dispatch executeScript target has no `allFrames` / `frameIds`
- [ ] Manual browser smoke (not run; CS handler covered by unit tests; activation regression covered by existing collision + eviction suites)

## Closes

- Closes #134
- Closes #135
- Closes #138

### DIFF
--- a/src/chrome/background/activation/__tests__/dispatch.test.ts
+++ b/src/chrome/background/activation/__tests__/dispatch.test.ts
@@ -195,6 +195,40 @@ describe('dispatchActivation', () => {
     expect(payload).toMatchObject({ type: 'activate-reader', scope: 'selection' });
   });
 
+  // Issue #135 — frame invariant on the contextMenu dispatch source.
+  // The happy-path command-source test pins target shape; this test
+  // pins the same invariant on a different source so a regression
+  // that adds `allFrames` to only one branch doesn't slip through.
+  it('#135: contextMenu source executeScript target has no allFrames / frameIds', async () => {
+    const { dispatchActivation } = await import('../dispatch');
+    const intent: ActivationIntent = {
+      source: 'contextMenu',
+      tabId: 121,
+      menuItemId: 'speedreader.ctx.preset.300.v1',
+    };
+
+    await dispatchActivation(intent);
+
+    const callArg = stub.scripting.executeScript.mock.calls[0]?.[0] as {
+      target: Record<string, unknown>;
+    };
+    expect(callArg.target).not.toHaveProperty('allFrames');
+    expect(callArg.target).not.toHaveProperty('frameIds');
+  });
+
+  // Issue #135 — same invariant on the popup dispatch source.
+  it('#135: popup source executeScript target has no allFrames / frameIds', async () => {
+    const { dispatchActivation } = await import('../dispatch');
+
+    await dispatchActivation({ source: 'popup', tabId: 122 });
+
+    const callArg = stub.scripting.executeScript.mock.calls[0]?.[0] as {
+      target: Record<string, unknown>;
+    };
+    expect(callArg.target).not.toHaveProperty('allFrames');
+    expect(callArg.target).not.toHaveProperty('frameIds');
+  });
+
   it('treats contextMenu without selectionText as full-scope', async () => {
     const { dispatchActivation } = await import('../dispatch');
     const intent: ActivationIntent = {
--- a/src/chrome/background/activation/__tests__/dispatch.test.ts
+++ b/src/chrome/background/activation/__tests__/dispatch.test.ts
@@ -69,11 +69,18 @@ describe('dispatchActivation', () => {
     expect(stub.tabs.get).toHaveBeenCalledWith(42);
     expect(stub.scripting.executeScript).toHaveBeenCalledTimes(1);
     const callArg = stub.scripting.executeScript.mock.calls[0]?.[0] as {
-      target: { tabId: number };
+      target: { tabId: number; frameIds?: unknown; allFrames?: unknown };
       files: string[];
     };
     expect(callArg.target.tabId).toBe(42);
     expect(callArg.files).toContain(CONTENT_SCRIPT_FILE);
+    // Issue #135 — top-frame-only invariant. Activation injects into
+    // the top frame; the dedup key is `(tabId, url)` keyed on the
+    // top-level document. Setting `allFrames: true` or `frameIds: [...]`
+    // here would silently break the dedup contract and let subframe
+    // origin flips bypass the `tab.url` recheck. Pin both.
+    expect(callArg.target).not.toHaveProperty('allFrames');
+    expect(callArg.target).not.toHaveProperty('frameIds');
     // The crxjs `?script` import resolves to the BUILT filename — must be
     // `.js`, not the `.ts` source. `chrome.scripting.executeScript` runs
     // against the emitted extension, not the source tree.
--- a/src/chrome/background/activation/__tests__/eviction.test.ts
+++ b/src/chrome/background/activation/__tests__/eviction.test.ts
@@ -331,6 +331,35 @@ describe('dispatchActivation — issue #128 lock eviction', () => {
     expect(rF.error.kind).toBe('inject-failed');
   });
 
+  // Review M3 — pin the behavior when executeScript REJECTS after the
+  // abort flag has been set. Current contract: the rejection wins (the
+  // catch branch returns the raw rejection's `details`) — the abort
+  // signal is redundant because the result is inject-failed either way.
+  // A regression that swallowed the rejection in favor of the abort
+  // sentinel would change the surfaced `details` and any caller
+  // branching on it would silently flip.
+  it('#138: executeScript rejection AFTER onRemoved — raw rejection details wins (abort signal is redundant)', async () => {
+    const { dispatchActivation } = await import('../dispatch');
+    const TAB = 403;
+    const intent: ActivationIntent = { source: 'command', tabId: TAB };
+
+    const pA = dispatchActivation(intent);
+    pending.push(pA);
+    await flushMicrotasks();
+
+    fireTabRemoved(TAB);
+    const rejectErr = new Error('Frame with ID 0 was removed');
+    executeCalls[0]?.reject(rejectErr);
+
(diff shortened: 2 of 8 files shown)


chriscantu:
## Swarm review applied

Four-critic antagonistic-ring (security, perf, scope, test-gap). Convergent findings actioned in commit fc26152; design-scope findings filed as follow-ups.

### Actioned (commit fc26152)

- **H1 + H2** [test-gap #1 + security #1]: CS listener wiring had zero tests AND no sender authorization. Added `sender.id === chrome.runtime.id` check + new `content/__tests__/listener.test.ts` (6 tests: registration, ok response, restricted-cs response, foreign-sender ignored, wrong-type ignored, non-object ignored).
- **M1** [test-gap #3]: frame invariant assertion was command-source only. Added popup + contextMenu source assertions in dispatch.test.ts.
- **M2** [security #3 + scope #1 + perf #3]: dropped `null as unknown as Promise<void>` placeholder cast. Builds `promise` first, then entry literal in one shot.
- **M3** [test-gap #2]: pinned executeScript rejection-after-abort contract — raw rejection details propagate, abort signal is redundant when result is already inject-failed.
- **L1** [scope #4]: introduced exported `INJECT_FAILED_TAB_REMOVED` constant for the `'tab-removed'` sentinel.

### Filed as follow-ups

- **H3** [test-gap #4]: slot-replace-then-onRemoved orphans leader's entry abort flag → #141. Needs design pass (track all in-flight entries per tabId vs slot-identity check at settle).
- **L5** [perf #1]: CS listener fan-out cliff at hundreds of tabs → #142.

### Considered, not actioned

- **M4** [scope #2]: `activate-handler.ts` split. Defensible — keeps the pure predicate unit-testable in isolation from the CS message-channel glue.
- **L2** [security #2]: handler returns `ok:true` while doing nothing. No SW caller branches on it today; flagged for the overlay-landing PR.
- **L3, L4, L6**: comment churn, bundle size, telemetry — cosmetic or below noise floor.

### Verification

```
npm run lint              # clean
npm run format:check      # clean
npm test                  # 507/507 (was 498; +9 review tests)
npm run build             # clean
npm audit --omit=dev --audit-level=high   # 0 vulnerabilities
```

Issue #173, opened 2026-05-30
Source OpenDyslexic-Regular.woff2 + SHA-256 pin + OFL.txt + CI integrity guard


chriscantu:
## Context

PR #169 shipped fonts WAR scaffold only — the `OpenDyslexic-Regular.woff2` binary is not committed (sandbox lacked web access). Ring security-adversary flagged supply-chain TOFU on the eventual binary commit: no pinned hash, no signature, no SRI. Whoever lands the file can swap a malicious / altered binary; OTS parser bugs (CVE-2023-4863-class) become an attack vector.

## Scope

Source + commit the binary with integrity gates:

1. Fetch `OpenDyslexic-Regular.woff2` from one of:
   - https://github.com/antijingoist/opendyslexic (upstream)
   - https://opendyslexic.org
2. Convert to WOFF2 if upstream ships TTF/OTF (`woff2_compress`)
3. Commit at `fonts/OpenDyslexic-Regular.woff2`
4. Commit `fonts/OFL.txt` (SIL OFL 1.1 license text — required by OFL §4)
5. Document the SHA-256 in `fonts/README.md`:
   ```
   OpenDyslexic-Regular.woff2  sha256:<hash>  source:<upstream-commit-sha>
   ```
6. Add a CI guard (`bin/verify-font-integrity.sh` or vitest) that:
   - Asserts `shasum -a 256 fonts/*.woff2` matches the pinned hashes in README
   - Asserts every `fonts/*.woff2` has an `OFL.txt` or `LICENSE` sibling

## Acceptance

- [ ] `fonts/OpenDyslexic-Regular.woff2` committed (real binary, file >10KB)
- [ ] `fonts/OFL.txt` committed
- [ ] `fonts/README.md` lists SHA-256 + upstream provenance
- [ ] CI guard validates hash + license-text presence
- [ ] `npm run build` succeeds; `dist/fonts/OpenDyslexic-Regular.woff2` byte-identical to source

## Blocks

#27 (OpenDyslexic toggle) and #28 (font picker) cannot ship a working OpenDyslexic experience until the binary lands.

## References

- PR #169 ring security-adversary finding #2 + finding #4
- PR #169 builder self-flag: "binary not committed"

Issue #172, opened 2026-05-30
Adopt use_dynamic_url:true on fonts WAR (privacy hardening)


chriscantu:
## Context

PR #169 fixed the broken `web_accessible_resources.matches: []` for bundled fonts by switching to `matches: ['<all_urls>']`. Ring security-adversary flagged a privacy concern: this exposes the extension to **cross-extension fingerprinting** — any web page can probe `chrome-extension://<id>/fonts/*` and detect SpeedReader installation.

## Proposal

Add `use_dynamic_url: true` to the fonts WAR entry in `src/chrome/manifest.ts`. Chrome rotates a per-session GUID on dynamic-URL WAR entries, blocking enumeration from arbitrary origins.

```ts
web_accessible_resources: [
  {
    resources: ['fonts/*'],
    matches: ['<all_urls>'],
    use_dynamic_url: true,  // <-- add
  },
],
```

## Trade-offs

- **Pro:** Blocks installation-fingerprinting + targeted-exploit surface.
- **Con:** `chrome.runtime.getURL('fonts/...')` returns a rotating URL — overlay code must NOT cache the URL across sessions. Verify the overlay font-injection path doesn't persist URLs to `chrome.storage`.

## Acceptance

- [ ] `use_dynamic_url: true` set on the fonts WAR entry
- [ ] Verify overlay font injection (will land via #27/#28) reads `chrome.runtime.getURL` fresh each mount
- [ ] Update `src/chrome/__tests__/manifest.test.ts` to assert `use_dynamic_url: true`
- [ ] Probe: from a webpage console, attempt `fetch('chrome-extension://<id>/fonts/...')` — should fail to resolve cross-origin without the session GUID

## References

- PR #169 ring security-adversary finding #1
- MV3 dynamic URL docs: https://developer.chrome.com/docs/extensions/reference/manifest/web-accessible-resources

PR #192, opened 2026-05-30
test(e2e): runtime CSP-strict smoke for bundled font loading (#174)

### LINKED ISSUE #174: e2e CSP-strict smoke test for bundled font loading via chrome-extension://
## Context

PR #169 wired fonts WAR plumbing but its builder flagged the CSP-strict runtime check as unverified. Ring test-gap-adversary surfaced the same gap: the PR claims "WAR plumbing wired" but plumbing that 404s under strict page CSP is not wired. Default extension CSP permits `chrome-extension:` font-src, but no automated check exists.

## Scope

Add a Playwright e2e smoke test that:
1. Serves a fixture page with a strict CSP:
   ```
   Content-Security-Policy: default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'
   ```
   (Notably: no `font-src` directive — `default-src 'none'` should block all font loads from the page origin)
2. Loads the SpeedReader unpacked extension
3. Activates the overlay on the fixture page
4. Injects a `@font-face` rule referencing `chrome.runtime.getURL('fonts/OpenDyslexic-Regular.woff2')`
5. Asserts:
   - No CSP violation reports in the page console
   - `document.fonts.check('1em OpenDyslexic')` resolves `true`
   - The overlay word region renders with the OpenDyslexic font (visual check via screenshot diff against system-font baseline)

## Dependency

Requires the issue tracking OpenDyslexic binary sourcing (see follow-up issue for SHA-256 pin) to land first — without a real `.woff2` in `fonts/`, the test can't verify load behavior.

## Acceptance

- [ ] `tests/e2e/fonts-csp.spec.ts` exists and runs in CI
- [ ] Test uses a strict-CSP fixture page that would block page-origin font loads
- [ ] Extension-injected `@font-face` loads successfully (proves `chrome-extension://` bypasses page CSP for WAR-declared resources)
- [ ] Test runs on the `test:e2e` CI job

## References

- PR #169 ring test-gap-adversary finding #5
- PR #169 builder self-flag: CSP-strict test unverified
- Existing Playwright suite: `tests/e2e/`

chriscantu:
## Summary

- Add `tests/e2e/fonts-csp.spec.ts` — Playwright e2e smoke that loads the unpacked extension on a fixture page served with `Content-Security-Policy: default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'` (no `font-src` directive, so `default-src 'none'` governs font fetches), mounts the overlay, and asserts the bundled `OpenDyslexic-Regular.woff2` loads via `chrome-extension://` without firing a CSP violation.
- Add `tests/e2e/fixtures/csp-strict.html` — minimal fixture page (no inline `<style>`, `<script>`, or font refs) so any CSP violation must originate from the overlay (the thing under test).
- Two new files; zero modifications to existing files (no harness refactor, no new deps).

Closes #174.

## What the spec contract proves

When a host page declares strict CSP without `font-src`:

1. The SpeedReader overlay mounts (production core-overlay bundle)
2. The bundled `OpenDyslexic-Regular.woff2` fetches successfully from `chrome-extension://<dynamic-id>/fonts/...`
3. No CSP-violation console messages fire
4. The overlay's `.word-region` resolves `'OpenDyslexic'` at the head of its computed font-family stack

This closes the runtime-check gap PR #169 wired without verifying. Adjacent infrastructure already landed: #172 (`use_dynamic_url: true`) + #173 (real binary + integrity guard).

## Mutation evidence (assertion discrimination)

| Mutation | Result | Why |
|---|---|---|
| Point FontFace probe at non-WAR `manifest.json` | **Test FAILS** (`A network error occurred.`) | Asserts WAR plumbing — catches a future PR dropping `fonts/*` from `web_accessible_resources` |
| Tighten CSP with explicit `font-src 'self'` | Test passes | Documented inline: Chromium exempts `chrome-extension:` WAR resources from page CSP `font-src` — assertion is a behaviour-change canary if Chrome ever tightens |
| `document.fonts.check('1em OpenDyslexic')` (rejected during authoring) | Returns true even with no @font-face registered (CSS Font Loading §4.2.3.6 fallback rule) — tautology trap | Replaced with FontFace constructor probe |

## Harness friction flagged

- **`document.fonts.check()` is non-discriminating** for absent faces — documented inline so future maintainers don't reintroduce the tautology.
- **Shadow-root `@font-face` rules do not propagate to `document.fonts`** in Chromium (verified empirically). FontFace probe sidesteps this by issuing the fetch from page scope against the same WAR URL — sufficient condition for the overlay's shadow-root @font-face to succeed under the same policy.
- **`use_dynamic_url: true` (#172) returns UUID-format IDs** not the static `[a-p]+` extension-id format. SW URL helper kept `[a-p]+` (correct — SW URLs always use the static ID); spec widened regex when reading the rotated WAR URL.
- **Strict CSP delivered via `context.route` interception** (header-attached, not `<meta>`) to match the production case for sites that matter (MDN, GitHub).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 916 pass (65 files)
- [x] `npm run build` — 205ms
- [x] `npm run verify:fonts` — ok
- [x] `npm run build:e2e-bundle` — 51ms
- [x] `npm run test:e2e` — 23 passed / 11 skipped (env-gated)
- [x] Targeted: `npx playwright test tests/e2e/fonts-csp.spec.ts` — 1 passed (190ms)
- [x] Mutation evidence — non-WAR target flips assertion red; restored

## References

- PR #169 ring test-gap-adversary finding #5
- PR #169 builder self-flag: CSP-strict test unverified
- #172 (`use_dynamic_url: true`) + #173 (real binary + integrity guard) — landed prerequisites

### DIFF
--- a/tests/e2e/fixtures/csp-strict.html
+++ b/tests/e2e/fixtures/csp-strict.html
@@ -1,0 +1,34 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>CSP-Strict Fixture (#174)</title>
+    <!--
+      The strict CSP header is supplied at HTTP-response level via
+      Playwright's `context.route` interception in tests/e2e/fonts-csp.spec.ts.
+      A meta http-equiv CSP would weaken the test — header-supplied CSPs are
+      the production case for sites like MDN, GitHub, banking apps. This
+      page intentionally declares no <style>, <script>, or font references
+      so that:
+
+      - If a CSP-violation report fires, it must originate from the
+        overlay's injected stylesheet / @font-face / shadow tree (the
+        thing under test), not from the page itself.
+      - If `document.fonts.check('1em OpenDyslexic')` returns true, the
+        font face MUST have been registered by the overlay (no page-side
+        registration exists).
+    -->
+  </head>
+  <body>
+    <main id="page-content">
+      <h1>CSP-strict fixture</h1>
+      <p>
+        This page exists to host the SpeedReader overlay under a
+        <code>default-src 'none'</code> Content Security Policy and verify
+        the bundled OpenDyslexic font loads from
+        <code>chrome-extension://</code> without CSP violations.
+      </p>
+    </main>
+  </body>
+</html>
--- a/tests/e2e/fonts-csp.spec.ts
+++ b/tests/e2e/fonts-csp.spec.ts
@@ -19,18 +19,56 @@
  * could block the font even though the manifest WAR plumbing is correct.
  * No automated test covered that path before this spec.
  *
- * Discriminating signal (mutation-tested during authoring)
- * --------------------------------------------------------
- * The "no CSP-violation console messages" + "document.fonts.check passes"
- * pair is the load-bearing assertion. If we deleted the `fonts/*` entry
- * from `web_accessible_resources` and rebuilt, the font fetch would
- * surface as a `Refused to load the font 'chrome-extension://…'` console
- * error from Chromium AND `document.fonts.check('1em OpenDyslexic')`
- * would resolve `false`. Computed-style assertion alone is NOT
- * discriminating — `font-family: 'OpenDyslexic', system-ui, …` resolves
- * the same string regardless of whether the font actually loaded. The
- * `document.fonts` check is what proves the woff2 binary was fetched
- * and parsed successfully under the strict CSP.
+ * Why CDP, not console string-matching
+ * ------------------------------------
+ * The first cut of this spec scraped console messages for English
+ * fingerprints ("refused to load", "violated", "content security
+ * policy"). Chromium has rephrased CSP messages historically — a wording
+ * change would silently empty the violation array and the test would
+ * pass vacuously. We now subscribe to two CDP signals:
+ *
+ *   - `Log.entryAdded` with `source: 'security'` — Chromium emits CSP
+ *     violations here in a structured form, independent of console text.
+ *   - `Network.requestWillBeSent` — proves the overlay's @font-face
+ *     rule actually issued a network request for the woff2 (F4(b) from
+ *     the ring-critic findings: a parallel FontFace probe could pass
+ *     while the overlay path was silently broken).
+ *
+ * Console scraping is retained as a belt-and-braces backup but is no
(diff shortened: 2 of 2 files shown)


chriscantu:
Folded all 5 tier-1 ring fixes into dc36db2 (single-file diff to `tests/e2e/fonts-csp.spec.ts`, +242/-90):

- **F1** — CSP **delivery** asserted (header equals `STRICT_CSP`) AND CSP **enforcement** asserted via positive control (cross-origin image load under `default-src 'none'` MUST fire ≥1 CDP `Log.entryAdded` security event). Without this, `context.route` silent-failure would let the test pass vacuously.
- **F2** — `routeHits` counter asserted `> 0` after navigation; `response.ok()` guard raises on upstream 5xx instead of empty-body silent pass.
- **F3** — Font-family regex replaced with split-on-`,` + index-0 assertion. Old pattern accepted `system-ui, OpenDyslexic, sans-serif` (fallback position); new pattern requires OpenDyslexic at index 0.
- **F4** — TWO discriminating signals: (a) shadow-root `@font-face` rule's `src` URL inspected directly, (b) CDP `Network.requestWillBeSent` counts requests to `fontUrl`. Closes the gap where a silently-broken overlay (e.g., `openDyslexicFontUrl` not threaded through) would still pass the page-scope FontFace probe.
- **F5** — Primary CSP-violation channel is now CDP `Log.entryAdded` filtered on `source: 'security'` (structured events, not console scraping). Console string-matching retained as belt-and-braces.

**Mutation evidence** (run during authoring):
- F1: stubbed `STRICT_CSP` to permissive policy → positive-control assertion fired red with `Received: 0` security events. Reverted.
- F4: dropped `openDyslexicFontUrl` prop from overlay → shadow-root `@font-face` check fired red first (upstream cheap check). Reverted.

**Verify gate** (all green locally): lint, format:check, 916/916 tests, build, verify:fonts, build:e2e-bundle, test:e2e (23 passed / 11 skipped). Targeted: 3x `npx playwright test tests/e2e/fonts-csp.spec.ts` at 445/454/444ms — tight variance, no flake.

**Harness note**: positive-control image load needs ~250ms `waitForTimeout` for CDP delivery (CSP refusal is synchronous in Blink but the CDP transport is async). 3x runs show comfortable margin.

Issue #49, opened 2026-04-19
Reading history


chriscantu:
Keep a list of recently-read articles with timestamps and resume state. Accessible from the popup. Respect privacy: opt-in, clearable.

Issue #25, opened 2026-04-19
Close (Esc + ✕)


chriscantu:
Esc key and ✕ button both close the overlay and restore the page. Preserve reading position in memory for the session (see future issue #48 for persistence).

PR #195, opened 2026-05-31
feat(reading-position): persist + restore per-URL with LRU cap (#48)

### LINKED ISSUE #48: Reading-position memory
Persist reading position per-URL so closing and reopening SpeedReader resumes where the user left off. Bound by storage quota — use `chrome.storage.local` with an LRU cap. User-facing: "resume" vs "start over" prompt on reopen.

chriscantu:
## Summary

- Persists RSVP reading position per canonical URL via `chrome.storage.local` with a 100-entry LRU cap.
- Auto-resumes the saved word index on overlay mount + shows a non-blocking `role="status" aria-live="polite"` toast offering a Start Over link (5 s auto-dismiss).
- Adapter-agnostic core (`src/core/storage/reading-position.ts`) + thin Chrome adapter (`src/chrome/storage/chrome-position-store.ts`). `src/core/**` boundary preserved: no `chrome.*` imports.

Closes #48 — first of M2 "v1 remaining" batch A (#48 storage layer → #49 history view consumes the same canonical-URL keyspace).

## Locked design decisions (none re-litigated mid-build)

| Axis | Decision | Why |
|---|---|---|
| URL canonicalization | Strip fragment + `utm_*`, lowercase host, preserve other query params | Same article reachable via tracking-tagged links should resume |
| Position granularity | Word index (`number`, 0-based) | RSVP renders per-word; natural restore unit; survives minor text drift |
| Restore UX | Auto-resume + dismissable toast | Industry default (Pocket, Instapaper); zero-click for users who want it |
| LRU cap | 100 URLs, evict oldest by `lastReadAt`, touch on every write | Bounded local storage; survives chrome.storage.local quotas |
| Onboarding interaction | None | New users have no positions; restore path is a no-op |

## Decisions beyond the locked five (surfaced for review)

1. **Persistence is full-scope only.** Selection-scope (`#25`) has no replay surface — selection text doesn't survive navigation. The session-scope shim still handles in-tab selection resume.
2. **1 s trailing-edge debounce on writes** + flush-on-close. At 600 wpm undebounced = 600 writes/min; the 1 s staleness window is invisible during reading.
3. **Resume precedence: session > persistent.** Same-session reopen is more current than a previous-session persisted value.
4. **`chrome.storage.local` availability guard** in the adapter — silently degrades on harnesses with a partial `chrome` namespace.

## Mutation evidence (assertion discrimination)

| Production line mutated | Mutation | Test that caught it |
|---|---|---|
| `canonicalize.ts:43` `key.startsWith(UTM_PREFIX)` | → `key.includes(UTM_PREFIX)` | `preserves params whose names contain but do not start with utm_` |
| `reading-position.ts:168` `while (filtered.length > POSITION_LRU_MAX)` | → `>=` (over-eviction) | 3 LRU tests went red (cap, overwrite-doesn't-promote, touch-protects-oldest) |
| `overlay.ts` toast guard `resumedAt !== null && opts.resumeToast` | dropped left clause | `mount with resumeToast but no real resume (initialIndex=0) does NOT render the toast` |
| `chrome-position-store.ts` adapter `get` | returned `{}` instead of delegating | `round-trips a write through chrome.storage.local` |

All mutations reverted; 979/979 still green after re-run.

## Files changed

**New (7):** `src/core/url/canonicalize.ts` + test, `src/core/storage/reading-position.ts` + test, `src/chrome/storage/chrome-position-store.ts` + test, `src/core/overlay/__tests__/reading-position-restore.test.ts`.

**Modified (6):** `src/chrome/content/index.ts` (store singleton + debounced writer + restore wiring), `src/core/overlay/overlay.ts` (toast DOM + auto-dismiss + onWordAdvance emission), `src/core/overlay/types.ts` (OverlayOptions extension), `src/core/overlay/constants.ts` (toast constants), `src/core/overlay/styles.ts` (`.resume-toast` + forced-colors block), `vitest.config.ts` (chrome/storage test glob).

## Test plan

- [x] `npm run lint` — clean (0 warnings, `--max-warnings=0`)
- [x] `npm run format:check` — all formatted
- [x] `npm test` — 979 passed / 979 (+39 vs base)
- [x] `npm run build` — clean, no TS errors
- [ ] Manual smoke: load unpacked dist/, read an article past word 50, close + reopen → toast shows "Resumed at word N of M", Start Over resets to 0
- [ ] Manual smoke: read 101 distinct articles → first article's position is gone, last 100 preserved (verify via DevTools → Application → Storage → Extension)

## Deferred follow-ups (not in this PR)

- Cross-device sync via `chrome.storage.sync` (quota constraints make it a redesign, not a parameter swap)
- History view consumer of `store.list()` — covered by #49
- URL-pattern allow/deny list ("never remember this site")
- Per-domain disable toggle in options
- Migration scaffold when `POSITION_SCHEMA_VERSION` bumps past 1

### DIFF
--- a/STRUCTURE.md
+++ b/STRUCTURE.md
@@ -47,14 +47,17 @@ src/
 │   ├── tokenize/             # Pure text → RSVP token stream
 │   ├── orp/                  # Optimal Recognition Point (focus character)
 │   ├── rsvp-engine/          # Word emission cadence (WPM → setTimeout)
-│   └── settings/             # Zod schema + defaults + version migrations
+│   ├── settings/             # Zod schema + defaults + version migrations
+│   ├── storage/              # Adapter-agnostic stores (e.g. reading-position #48)
+│   └── url/                  # URL canonicalization (storage-key join key, #48)
 ├── chrome/                   # Chrome MV3-specific glue
 │   ├── manifest.ts           # Generates manifest.json at build time
 │   ├── background/           # Service worker entry
 │   ├── content/              # Content script entry (lazy-injected)
 │   ├── popup/                # Browser-action popup (HTML + TS)
 │   ├── options/              # Options page (HTML + TS)
-│   └── settings/             # chrome.storage.sync adapter over core/settings
+│   ├── settings/             # chrome.storage.sync adapter over core/settings
+│   └── storage/              # chrome.storage.* glue for core/storage stores (#48)
 └── icons/                    # (build-time icon staging, if used)
 ```
 
--- a/src/chrome/content/__tests__/position-flush.test.ts
+++ b/src/chrome/content/__tests__/position-flush.test.ts
@@ -95,7 +95,10 @@ afterEach(() => {
 });
 
 describe('content script — visibility/pagehide flush wiring (#48)', () => {
-  async function mountOverlayWithPendingWrite(): Promise<{ local: FakeStorageLocal }> {
+  async function mountOverlayWithPendingWrite(): Promise<{
+    local: FakeStorageLocal;
+    pageUrl: string;
+  }> {
     const { local, getListener } = installChromeStub();
     await import('../index');
     const listener = getListener();
@@ -111,12 +114,33 @@ describe('content script — visibility/pagehide flush wiring (#48)', () => {
     // engine. Fast-forward to confirm at least one advance fires:
     await new Promise((r) => setTimeout(r, 250));
 
-    return { local };
+    return { local, pageUrl: 'https://example.com/article-flush-test' };
   }
 
-  test('visibilitychange → hidden triggers a flush of the pending write', async () => {
-    const { local } = await mountOverlayWithPendingWrite();
+  /**
+   * Extracts the wordIndex written under the canonical `position:<url>`
+   * key from a `chrome.storage.local.set` call payload. The flush path
+   * writes the per-URL payload AND the LRU index; we only assert on
+   * the per-URL payload so the test is robust to index-key changes.
+   */
+  function payloadWordIndex(setArg: Record<string, unknown>, pageUrl: string): number | undefined {
+    const key = `position:${pageUrl}`;
+    const v = setArg[key];
+    if (v === null || typeof v !== 'object') return undefined;
+    const wi = (v as { wordIndex?: unknown }).wordIndex;
+    return typeof wi === 'number' ? wi : undefined;
+  }
+
+  test('visibilitychange → hidden flushes the pending write AND cancels the debounce timer', async () => {
+    // TG3 — spy clearTimeout BEFORE the import so the production code's
+    // calls are captured. `setTimeout` returns numeric IDs in jsdom;
+    // we capture the most-recently scheduled debounce timer ID and
+    // assert clearTimeout was called with it BEFORE local.set fires.
+    const clearTimeoutSpy = vi.spyOn(globalThis, 'clearTimeout');
+
+    const { local, pageUrl } = await mountOverlayWithPendingWrite();
     local.set.mockClear();
+    clearTimeoutSpy.mockClear();
 
     // Default visibilityState in jsdom is 'visible' — override it.
     Object.defineProperty(document, 'visibilityState', {
@@ -129,9 +153,21 @@ describe('content script — visibility/pagehide flush wiring (#48)', () => {
     // microtasks to settle.
     await new Promise((r) => setTimeout(r, 30));
 
-    // Expect at least one set call (position payload) — the flush
-    // path writes the position record AND the LRU index.
+    // (a) clearTimeout was called as part of the flush — proves the
+    // debounce timer was cancelled, not raced against.
+    expect(clearTimeoutSpy).toHaveBeenCalled();
+
+    // (b) The first set call carries the per-URL payload with a
+    // non-zero wordIndex (engine has been running ~250 ms at 600 wpm
+    // → ~2-3 word events) — proves the pending write actually
+    // flushed, not just any storage call.
     expect(local.set).toHaveBeenCalled();
+    const firstCall = local.set.mock.calls[0][0] as Record<string, unknown>;
+    const wi = payloadWordIndex(firstCall, pageUrl);
+    expect(wi).toBeDefined();
+    expect(wi).toBeGreaterThan(0);
+
+    clearTimeoutSpy.mockRestore();
   });
 
   test('visibilitychange → visible does NOT trigger a flush', async () => {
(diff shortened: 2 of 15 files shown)


chriscantu:
Folded ring fixes into `e45882f` — 5 HOLDs + 3 NITs landed; S3 deferred with reasoning.

**Surgical fixes (5 HOLDs):**

| Finding | Resolution |
|---|---|
| S1 (sec, scheme allow-list + length cap) | `canonicalizeUrl` returns `null` for non-`http(s)` schemes (`javascript:`, `data:`, `blob:`, `chrome*:`, `file:`, `about:`, `ws/wss`) and canonical forms >2048 chars. Store write/read silently no-op on null keys. Closes the quota-DoS via `history.pushState` spam + namespace pollution that would have broken `saveSettings`. |
| S2 (sec, LRU race) | `ReadingPositionStore` serializes `write`/`touch`/`clear` through a per-instance promise queue. 10-concurrent-writes test (deterministic adapter delay between get/set) confirms no orphaned payloads or dangled index entries. |
| A1 (arch, `list()` for #49) | Added `list()` returning entries in LRU order (most-recent first) + `clearAll()` (wipes index + all `position:*` keys). #49 can consume the interface without modifying it. |
| A2 (arch, schema-version forward-incompat) | `write()` pre-reads stored payload; aborts silently if `stored.schemaVersion > POSITION_SCHEMA_VERSION`. Preserves forward-compatible records on downgrade. |
| A3 (arch, flush on navigation) | Wired `flushPendingPositionWrite()` to `visibilitychange` (state === 'hidden') AND `pagehide`. Listeners detach on overlay close. Closes the most common close path (user navigates away with overlay open). |

**NITs folded:**
- STRUCTURE.md — added rows for `src/core/storage/` + `src/chrome/storage/`
- `types.ts` — JSDoc on `onWordAdvance` noting selection-scope exclusion
- `index.ts` — `console.warn` once on `chrome.storage.local` absence (matches `fontUrl` branch pattern), not silent-degrade

**Test delta:** 979 → **1022** (+43). Lint, format:check, build, all green.

**Mutation evidence (9 production lines mutated, each caught by ≥1 test):**

| Finding | Mutation | Caught by |
|---|---|---|
| S1 scheme | dropped scheme check | 13 scheme-allow-list tests went red |
| S1 length | dropped length cap | 1 length test red |
| S1 null no-op | leaked rejected key into adapter | 7 non-canonicalizable-skip tests red |
| S2 queue | bypassed enqueue | 10-concurrent-writes test: index length 1 instead of 10 |
| A1 list order | iterated oldest-first | touch-reorder test: wrong URL at index 0 |
| A1 clearAll | skipped removing index sentinel | snapshot length 1 |
| A2 schema guard | removed `isForwardIncompat` pre-check | v2 record overwritten by v1 write |
| A3 visibility | removed `addEventListener` | `local.set` never called |
| A3 detach | removed `detachPositionFlushListeners()` | `removeEventListener` never called for `visibilitychange` |

All mutations reverted; 1022/1022 still green.

---

**S3 (cross-context enumeration) — DEFERRED with structural reasoning.**

Builder's analysis (and I agree on review): both candidate key-storage paths defeat the threat model.

- **`chrome.storage.session` for AES key** — SW eviction (~30s idle) wipes the key, making encrypted records unreadable. Counterproductive for a persistence feature whose value proposition is bridging sessions. User would see "history vanished" every time the SW idle-evicted.
- **`chrome.storage.local` for AES key** — content scripts read the key alongside the encrypted blobs. SHA-256 URL hashing is a public algorithm with no obstruction value. Net security improvement vs plaintext: zero.

The real mitigation is **structural, not cryptographic**: move the position store into the service worker with a message-passing API, removing content-script `chrome.storage.local` access entirely. Encryption on top of that move is security theater; without that move, encryption doesn't help.

Filing as a separate spec PR — the synchronous-CS → async-SW-round-trip change deserves design review (SW cold-start race window, message-ordering during overlay mount, perf cost at 1Hz write rate) rather than a jam into this surgical-fixes PR.

S3 follow-up issue + spec PR to land before any S3 impl.

chriscantu:
Second adversarial ring pass surfaced 7 HOLDs (test-gap, type-design, code-reviewer, chrome-extension-engineer). Folded into `046c416`. NITs deferred.

**Round-2 fixes:**

| Finding | Resolution |
|---|---|
| TG1 (test-gap) | Write-queue serialization test was tautological — 10 distinct URLs = no race. Rewrote to use same-URL concurrent writes against a slow adapter; asserts last-write-wins + single index entry. S2 fix now actually verified. |
| TG2 (test-gap) | Queue-poisoning JSDoc claim untested. Added test where adapter first set() rejects → asserts (a) failing write rejects, (b) subsequent write resolves. Brief's mutation hypothesis was inaccurate; corrected to drop onRejected from HEAD enqueue. Validated. |
| TG3 (test-gap) | Pagehide flush asserted only `toHaveBeenCalled()`. Tightened: assert payload wordIndex matches latest pending write AND `clearTimeout` called before `local.set` (timer cancelled, not raced). |
| TG4 (test-gap) | Re-mount accumulation untested. Added mount → close → mount-with-different-URL → pagehide cycle. Honest disclosure: mutation discriminator lives in adjacent existing detach test (module-scope handlers mean only one pair in flight); new test is integration-level. |
| TG6 (test-gap) | `clearAll()` JSDoc/impl divergence. Decision: docs match impl (`get(null)` for orphan scan would load all extension storage including settings, too expensive). Rewrote JSDoc to document index-only sweep + orphan preservation; regression test pins behavior. Orphan-sweep follow-up to file. |
| TD1 (type-design) | Cross-field invariant `wordIndex < totalWords` type-invisible. Added module-private `makeReadingPosition()` smart constructor unifying validation across `parseStored` + `write`. Fixed 2 pre-existing LRU tests that violated the new invariant (bumped `totalWords` to preserve semantics under test). |
| TD6 (type-design) | `chromeStorageAdapter` JSDoc claimed "safe to reuse" but each `createReadingPositionStore` allocates fresh write-queue. Verified `content/index.ts:49` already constructs at module scope (load-bearing); fixed misleading docs to require single-construct per content-script lifetime. |

**Tests:** 1022 → **1038** (+16). Lint, format:check, build all green.

**Builder self-pushback (quality signal):** corrected 2 inaccurate mutation hypotheses in the brief via direct verification rather than fabricating coverage. TG2 real discriminator is `then(op, op)` vs `then(op)` on HEAD, not the tail recovery (self-heals via next enqueue). TG4 detach mutation is caught by adjacent existing test, not the new TG4 test — disclosed honestly.

**NITs deferred to follow-ups (13 total):**
- TG5 (utm-strip → under-cap edge case)
- TD2 (over-exported keyspace constants)
- TD3 (`onWordAdvance` arg-swap)
- TD4 (name `WritableReadingPosition`)
- TD5 (document `StorageAdapter` error contract)
- ME1 (chrome.alarms backstop doc)
- ME2 (minimum_chrome_version doc)
- ME3 (test fake doesn't catch callback-form regression)
- ME5 (mount/close asymmetric)
- ME6 (singleton injection-vs-module doc)
- ME7 (cross-tab serialization per-instance doc)
- CR1 (folded into TG1 rewrite — dead `%1===0` removed)
- CR2 (page word-count drift sanity guard)

**ME4 deferred to #196 spec:** `await store.write()` in pagehide handler won't survive SW migration; spec needs to decide between long-lived `connect()` ports vs `chrome.storage.session` as the unload-safe write path.

Two follow-up issues to file (coordinator):
1. `clearAll()` orphan-sweep maintenance task (TG6)
2. Bundle of remaining NITs by theme (test-fidelity, type-ergonomics, MV3-docs)

PR #201, opened 2026-05-31
feat(rsvp): chunk-stream emission for 2–3 word display (#51)

### LINKED ISSUE #51: Chunk size (2–3 word display)
Allow displaying 2 or 3 words at once instead of one. Options-page setting. Reuses ORP logic per-word. Useful for faster readers who want more context at a glance.

## Safari parity (sourced from `chunk-builder.test.js` gap analysis in #88)

The Safari `buildChunks(words, chunkSize)` helper has two load-bearing requirements that should land with this issue:

1. **Sentence-boundary respect** — a chunk never crosses sentence-end punctuation. If the configured chunkSize is 3 but a sentence ends after 2 words, the chunk closes at 2.
2. **Index tracking** — each chunk carries `startIndex` / `endIndex` into the underlying word array so seek / progress / scrubber surfaces (#47, #95, #96) can map back to positions.

## Dependencies

- **#90** — sentence-boundary detection. `buildChunks` cannot enforce requirement 1 without `sentenceStart` flags on the word stream.

## Safari reference

`chriscantu/speed-reader/tests/js/chunk-builder.test.js` — 11 cases covering chunkSize 1/2/3, odd-count handling, sentence-boundary respect, single-word sentences, and edge cases. Port those cases to vitest when this lands.

## Definition of Done (revised)

- [ ] `buildChunks` (or equivalent) implemented in `src/core/rsvp-engine/` (or a sibling `chunks/` module)
- [ ] Wired into `createRsvpEngine` so emissions can be word- or chunk-shaped based on a `chunkSize` option
- [ ] Settings schema bump for `chunkSize` (1, 2, or 3; default 1 — Safari default 1)
- [ ] Options-page surface (deferred to that issue if necessary)
- [ ] Sentence-boundary respect verified (depends on #90)
- [ ] 11 Safari `buildChunks` cases ported and passing

chriscantu:
# feat(rsvp): multi-word chunk display (1–3 words) — closes #51

Adds Safari-parity multi-word RSVP via a new `chunkSize` setting (1 | 2 | 3, default 1) and engine `chunk` event emission. Sentence boundaries are always respected — a chunk never spans a sentence start. Default behavior is unchanged for every existing user (chunkSize=1 → today's word emission path).

## Scope

- `src/core/rsvp-engine/chunks.ts` — pure `buildChunks(words: WordToken[], chunkSize: 1|2|3): Chunk[]` ported verbatim from Safari `chunk-builder.js` semantics.
- `src/core/rsvp-engine/rsvp-engine.ts` — additive `chunkSize?: ChunkSize` option; engine emits `chunk` events when `chunkSize >= 2`. Word mode untouched (line-for-line identical to pre-change behavior).
- `src/core/rsvp-engine/__tests__/chunks.test.ts` — 10 cases ported from Safari source (covers the 10 inner `it` blocks in chriscantu/speed-reader's chunk-builder tests; spec said "11" but the Safari source has 10 `it` blocks across the listed describe groups).
- `src/core/rsvp-engine/__tests__/chunk-mode.test.ts` — 16 engine integration cases covering emission, sentence boundaries, pacing, seekTo, seekToSentence, progress, back-compat.
- `src/core/settings/schema.ts` + `migrations.ts` + `defaults.ts` — V5 → V6 bump adding `chunkSize: 1|2|3`. Migrator stamps `chunkSize: 1` for every existing payload (back-compat default).
- `src/chrome/options/{controller.ts,index.html}` — add Words-per-display select under Appearance section.
- `src/chrome/content/index.ts` + `src/core/overlay/{types.ts,overlay.ts}` — wire `chunkSize` from settings → overlay → engine; overlay subscribes to and renders `chunk` events alongside `word` events.
- Mass type-rename `SettingsV5 → SettingsV6` across non-legacy consumers (storage, options, welcome, context-menu, content, overlay types, plus their tests). Required mechanically by the schema bump; existing V5-as-legacy tests retained their `SettingsSchemaV5` reference for migration-consumer coverage.

## Decisions on Open Questions

Per coordinator brief, these were proposed inline rather than silently chosen.

### Q1 — Seek semantics in chunk mode

**Decision: always snap-to-chunk-start.** The external `seekTo(index)` API still takes a word-axis index; internally the engine snaps to the chunk containing that word and emits the whole chunk.

**Rationale:** chunk mode makes the chunk the display unit. Allowing mid-chunk seeks would force the engine to either (a) emit a partial chunk straddling the requested word, breaking the chunk-as-displayable-unit contract, or (b) emit a different chunk than the next-tick chunk, breaking the advance invariant. Snap is the only choice that keeps both invariants intact.

**Trade-off accepted:** consumers that want sub-chunk precision (e.g., a hypothetical reading-position store keying to the word axis) get the chunk-start position back, not their requested position. Acceptable because the only current seek consumer (`#48` reading position restore) just needs "resume near here" semantics.

### Q2 — Punctuation pacing in chunk mode

**Decision: use the chunk's last-word multiplier.** The gap before the next chunk = `base * pacingMultiplier(lastWordOfJustEmittedChunk)`.

**Rationale:** the pause comes AFTER the displayed text. Perceptually, only the trailing punctuation of the displayed chunk governs how long the pause should feel. Sum-of-multipliers (1.2 × 1.5 = 1.8× for chunks containing both a clause break and a sentence-final) would over-pace — the user sees one screen of text and one pause; mathematically averaging multiple internal pauses adds time without perceptual basis.

**Counter considered:** max-of-multipliers (use the strongest pause-inducer in the chunk). Rejected — that's an opaque rule and produces the same result as last-word for the common case (sentence-final is always the strongest signal and almost always lands last in a chunk).

### Q3 — ORP highlighting inside multi-word chunks

**Decision: out of scope for this PR — follow-up issue.** The chunks engine emits a chunk's text; the overlay currently renders that text verbatim via `renderWord(word, ev.text)`. ORP highlighting is an overlay-rendering concern; per-word ORP within a chunk vs. single center-word ORP is a UX question that deserves its own design pass.

**Why split:** wedging ORP into this PR would balloon scope and force a snap-decision on a UX question that's not strictly required for chunk mode to be usable. A user opting into chunkSize=2 today gets a readable multi-word display without ORP markers; adding ORP later is an additive overlay-side change with no engine churn.

**Follow-up:** open a Chrome-side issue for "Multi-word chunk ORP rendering — center-word vs per-word". Repro: set chunkSize=2 in options, open SpeedReader on any article. Current display = chunk text rendered without ORP markers. Safari's overlay uses per-word ORP within chunks (visible in `Resources/rsvp/overlay-render.js`); Chrome's `core/overlay/word.ts` would need to walk chunk words and apply ORP per token.

### Q4 — When `chunkSize === 1`

**Decision: keep `word` event emission (back-compat).** `chunkSize` undefined OR `=== 1` → engine behaves exactly as today (emits `word` events). Only `chunkSize >= 2` enables chunk mode.

**Rationale:** preserves bit-for-bit back-compat for all 109 existing engine tests and every consumer reading `word` events today. The alternative (uniform `chunk` events for all chunkSize values) would either force a Big Bang overlay change (every existing user's display path rewires through chunk rendering) or require dual rendering paths anyway.

**Trade-off accepted:** the chunk-mode boolean is implicit ("chunkSize > 1") rather than explicit. The internal engine constructor normalizes this once into a single `chunkSize: ChunkSize | undefined` variable (`>= 2 ? options.chunkSize : undefined`), so consumer-side code doesn't see the distinction.

## Mutation-hypothesis table (self-pushback)

| Test assertion | Mutation that would silently pass without this test | Why caught |
|---|---|---|
| `buildChunks` `startIndex` / `endIndex` exact values | Off-by-one in chunk bounds (e.g. `i = j - 1` instead of `i = j`) | `chunks[0].endIndex === 1` for chunkSize=2, `chunks[1].startIndex === 2` — both indices pinned per chunk |
| `buildChunks` sentence-boundary break | `sentenceStart` flag ignored (chunkSize=3 over `Hello world. Goodbye world.` would produce one chunk) | Test asserts exactly 2 chunks at chunkSize=3 with explicit text per chunk |
| `chunkSize=3` regressing to `chunkSize=2` | `chunkWords.length < chunkSize` becomes `<= chunkSize - 1` (off-by-one cap) | `chunkSize=3` test asserts chunks of length 3 in `["The","quick","brown","fox","jumps","over."]` |
| V5→V6 migrator default | Migrator omits chunkSize stamp (relies on safeParse fallback) | Tests pin `result.chunkSize === 1` for V5 input AND for hand-edited bogus values |
| V5→V6 spread-order regression | Migrator flips to `{ chunkSize: 1, ...raw }` letting bogus user value through | `it.each` covers 6 invalid types — each asserts `result.chunkSize === 1` regardless |
| `word` event discriminator collision | Switching `word` to `chunk` for chunkSize=1 | Two back-compat tests (`chunkSize=1` and `chunkSize=undefined`) assert exact `{ type: 'word', ... }` shape |
| Punctuation pacing in chunk mode | Using first-word instead of last-word multiplier | Test asserts gap = 150ms for sentence-final-ending chunk vs 100ms base; first-word would yield 100ms (no period on "Hello") and fail |
| Chunk reference identity | `chunkWords.push({...words[j]})` (cloning instead of referencing) | Edge-case test asserts `chunks[0].words[0] === w[0]` with `toBe` (reference equality) |
| seekTo snap behavior | mid-chunk seek emits a synthetic single-word chunk instead of snapping | Test asserts the full chunk text `"brown fox."` and exact indices |
| seekToSentence forward boundary | Walk starts from `cur+1` instead of `cur`, OR uses sentenceStart-after instead of sentenceEnd-after | 3-sentence test pins exact text emitted (`"C D."`) after next-sentence from chunk 0 |
| Options chunkSize HTML drift | Reader accepts the rogue `<option value="4">` | Test injects a `value="4"` option, dispatches change, asserts saveMock was NOT called with chunkSize patch |

## Self-weaknesses

1. **ORP within chunks is unresolved** (Q3). The current overlay renders chunk text without ORP markers; Safari upstream does per-word ORP. Tracked as follow-up. Not a regression — single-word users (chunkSize=1, the default) keep the existing ORP behavior because they stay on the word-event path.
2. **Live chunkSize updates not wired.** `subscribeSettings` forwards `chunkSize` to OverlaySettings but the overlay does NOT reconstruct the engine on chunkSize change — only on next mount. Documented in `OverlaySettings.chunkSize` JSDoc. Adding live updates would require an engine swap or new public `engine.setChunkSize()` method; deferred to keep scope tight.
3. **Word-mode engine axis vs chunk-mode axis divergence in `progress()`.** Word mode: `total = words.length` (raw, including paragraph sentinels and dashes). Chunk mode: `total = filteredWords.length` (only `kind:'word'` tokens). For an article with 5% structural tokens, the two would report different totals. Acceptable because progress is presented to humans as a percentage and the user perceives "words consumed", not "tokens consumed"; but a consumer that depends on absolute counts across mode-switches would see a discontinuity. The settings-storage code does not currently cross-reference these counts.
4. **`seekToSentence` semantics in chunk mode are an interpretation, not a Safari port.** Safari's seekToSentence operates on its word-with-index model; my chunk-mode adaptation walks the marker flags using the chunk-just-emitted as the anchor. I think the semantics are correct (chunk 0 emitted "A B." → next-sentence should play "C D." which contains the next sentence), but if Safari's UX differs in subtle ways for chunk mode, this is the place to look.
5. **Non-word tokens silently dropped in chunk mode.** Paragraph sentinels and em/en dashes never appear in chunks (filtered out before `buildChunks`). Word mode emits them as-is. Probably MORE correct than today's "emit `\n\n` as a word" behavior, but it IS a behavioral difference for power users that turn on chunk mode.
6. **Test count discrepancy with brief.** Brief says "11 ported Safari `buildChunks` test cases verbatim". The Safari source has 10 `it` blocks across the listed describe groups (1 + 2 + 1 + 3 + 3 = 10). I ported all 10 verbatim. If there's an 11th I missed, please point me at it.

## CI output (tails)

```
$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
```
(clean; no warnings, no errors)

```
$ npm run format:check
Checking formatting...
```

```
$ npm test
 Test Files  76 passed (76)
      Tests  1177 passed (1177)
   Start at  00:26:07
   Duration  3.14s
```

```
$ npm run build
dist/assets/index.ts-Mp61JVsW.js                 33.48 kB │ gzip: 11.23 kB
✓ built in 221ms
```

```
$ npx tsc --noEmit
(clean — no output)
```

## Test plan

- [x] `npm run lint` clean
- [x] `npm run format:check` clean
- [x] `npm test` — 1177 passed (8 new tests in this PR: 10 chunks + 16 chunk-mode + 9 new schema + 7 new migrator + 5 new controller minus overlap with existing assertions)
- [x] `npm run build` clean
- [x] `npx tsc --noEmit` clean
- [x] ~Manual smoke: load extension, open options, switch chunkSize to 2, open SpeedReader on a long article, confirm multi-word chunks display with sentence-boundary respect~ — **automated** in `tests/e2e/chunk-mode.spec.ts` (`chunkSize=2 — pairs, never crossing sentence boundaries`)
- [x] ~Manual smoke: switch chunkSize to 3, confirm 3-word chunks with shorter chunks at sentence boundaries~ — **automated** in `tests/e2e/chunk-mode.spec.ts` (`chunkSize=3 — triples, sentence boundaries shorten chunks`)
- [x] ~Manual smoke: switch chunkSize back to 1, confirm single-word display restored~ — **automated** in `tests/e2e/chunk-mode.spec.ts` (`chunkSize=1 — single-word display per emission`)
- [x] ~Manual smoke: opt into chunkSize=2 with punctuation pacing on, confirm longer pauses after sentence-ending chunks~ — **covered by unit tests** in `src/core/rsvp-engine/__tests__/chunk-mode.test.ts` ("punctuation pacing (chunk last-word multiplier)" describe block). Virtual-timer unit tests give deterministic ms-delta assertions; in-browser timing would be flaky.

### Test plan execution log

E2e suite: 26 passed, 11 skipped (pre-existing skips unrelated to this PR), 0 failed.
Unit suite: 1197 passed, 0 failed.
Lint / format:check / tsc / build: all clean.

### DIFF
--- a/src/chrome/background/context-menu/__tests__/factory.test.ts
+++ b/src/chrome/background/context-menu/__tests__/factory.test.ts
@@ -6,13 +6,13 @@
  */
 
 import { describe, it, expect } from 'vitest';
-import type { SettingsV5 } from '../../../../core/settings';
+import type { SettingsV6 } from '../../../../core/settings';
 import { DEFAULT_SETTINGS } from '../../../../core/settings/defaults';
 import { buildMenuItems, type CtxMenuItemId } from '../factory';
 
 const HTTP = ['http://*/*', 'https://*/*'];
 
-function withSettings(partial: Partial<SettingsV5>): SettingsV5 {
+function withSettings(partial: Partial<SettingsV6>): SettingsV6 {
   return { ...DEFAULT_SETTINGS, ...partial };
 }
 
--- a/src/chrome/background/context-menu/__tests__/install.test.ts
+++ b/src/chrome/background/context-menu/__tests__/install.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach, vi, type Mock } from 'vitest';
-import type { SettingsV5 } from '../../../../core/settings';
+import type { SettingsV6 } from '../../../../core/settings';
 import { DEFAULT_SETTINGS } from '../../../../core/settings/defaults';
 
 vi.mock('../../../settings/storage', () => ({
@@ -101,7 +101,7 @@ describe('ensureContextMenu', () => {
     stub.contextMenus.create.mockClear();
     stub.contextMenus.update.mockClear();
 
-    const nextSettings: SettingsV5 = { ...DEFAULT_SETTINGS, lastUsedWpm: 420 };
+    const nextSettings: SettingsV6 = { ...DEFAULT_SETTINGS, lastUsedWpm: 420 };
     loadSettings.mockResolvedValueOnce(nextSettings);
     await ensureContextMenu();
 
@@ -122,7 +122,7 @@ describe('ensureContextMenu', () => {
     stub.contextMenus.create.mockClear();
     stub.contextMenus.update.mockClear();
 
-    const nextSettings: SettingsV5 = { ...DEFAULT_SETTINGS, contextLine: true };
+    const nextSettings: SettingsV6 = { ...DEFAULT_SETTINGS, contextLine: true };
     loadSettings.mockResolvedValueOnce(nextSettings);
     await ensureContextMenu();
 
@@ -155,7 +155,7 @@ describe('ensureContextMenu', () => {
     let loadResolved = false;
     loadSettings.mockImplementationOnce(
       () =>
-        new Promise<SettingsV5>((resolve) => {
+        new Promise<SettingsV6>((resolve) => {
           setTimeout(() => {
             loadResolved = true;
             resolve(DEFAULT_SETTINGS);
--- a/src/chrome/background/context-menu/factory.ts
+++ b/src/chrome/background/context-menu/factory.ts
@@ -4,7 +4,7 @@
  * No `chrome.*` calls — `install.ts` owns the adapter side. Keeping this
  * file pure means every shape decision (titles, checkbox state, separator
  * placement) is exercised by `factory.test.ts` against a plain
- * `SettingsV5` input without any chrome-stub scaffolding.
+ * `SettingsV6` input without any chrome-stub scaffolding.
  *
  * See:
  * - `docs/superpowers/specs/2026-05-25-context-menu-integration.md`
@@ -13,7 +13,7 @@
  *   §"File Layout" — the factory / adapter / listener split.
  */
 
-import type { SettingsV5 } from '../../../core/settings';
+import type { SettingsV6 } from '../../../core/settings';
 
 /**
  * Stable, versioned menu-item IDs. The `v1` suffix lets a future shape
@@ -63,7 +63,7 @@ const HTTP_PATTERNS = ['http://*/*', 'https://*/*'] as const;
  * `chrome.*` references. Items are emitted in the order they should
  * appear in the submenu; the adapter preserves order on `create`.
  */
-export function buildMenuItems(settings: SettingsV5): CtxItemSpec[] {
+export function buildMenuItems(settings: SettingsV6): CtxItemSpec[] {
   const PARENT_ID: CtxMenuItemId = 'speedreader.ctx.parent.v1';
 
   const parent: CtxItemSpec = {
--- a/src/chrome/content/__tests__/chunk-size-wire.test.ts
+++ b/src/chrome/content/__tests__/chunk-size-wire.test.ts
@@ -1,0 +1,198 @@
+/**
+ * chunk-size-wire.test.ts — issue #51 (test-gap HIGH #3)
+ *
+ * End-to-end wire from `chrome.storage.sync.get` → loadSettings →
+ * createOverlay (`initialSettings.chunkSize` slot) → subscribeSettings
+ * push → `engine.setChunkSize(next)`. The unit tests in
+ * `core/rsvp-engine/__tests__/chunk-mode.test.ts` pin the engine's
+ * local behaviour; this file pins the chrome-glue boundary so a
+ * regression that drops `chunkSize: settings.chunkSize` from the CS
+ * payload or the subscribe wiring lands as a test failure rather
+ * than a silent default at the boundary.
+ *
+ * Mutation guards:
+ *   - Hardcoding `chunkSize: 1` at the initial-mount site must fail
+ *     the first test (which loads chunkSize=3 and asserts chunk
+ *     events are emitted by the underlying engine).
+ *   - Dropping `chunkSize: s.chunkSize` from the subscribe wiring
+ *     must fail the second test (settings push to chunkSize=2
+ *     should switch the live event stream from word to chunk).
+ */
+import { describe, expect, test, vi, beforeEach, afterEach } from 'vitest';
+import { DEFAULT_SETTINGS } from '../../../core/settings/defaults';
+import type { SettingsV6 } from '../../../core/settings/schema';
+import { OVERLAY_CLASS } from '../../../core/overlay/constants';
+
+const SETTINGS_KEY = 'speedreader.settings';
+const EXT_ID = 'test-ext';
+
+type Listener = (
+  msg: unknown,
+  sender: { id?: string },
+  sendResponse: (r: unknown) => void,
+) => unknown;
+
+type StorageChangedListener = (
+  changes: Record<string, { newValue?: unknown; oldValue?: unknown }>,
+  area: string,
+) => void;
+
+interface ChromeMock {
+  runtime: {
+    id: string;
+    onMessage: { addListener: (l: Listener) => void };
+    getURL: ReturnType<typeof vi.fn>;
+  };
+  storage: {
+    sync: {
+      get: ReturnType<typeof vi.fn>;
+      set: ReturnType<typeof vi.fn>;
+    };
+    onChanged: {
+      addListener: ReturnType<typeof vi.fn>;
+      removeListener: ReturnType<typeof vi.fn>;
+    };
+  };
+}
+
+function installChromeMock(stored: SettingsV6): {
+  mock: ChromeMock;
+  getListener: () => Listener;
+  emitStorageChange: (next: SettingsV6) => void;
+} {
+  let capturedListener: Listener | undefined;
+  const storageListeners: StorageChangedListener[] = [];
+  const mock: ChromeMock = {
+    runtime: {
+      id: EXT_ID,
+      onMessage: {
+        addListener: (l: Listener) => {
+          capturedListener = l;
+        },
+      },
+      getURL: vi.fn((path: string) => `chrome-extension://${EXT_ID}/${path}`),
+    },
+    storage: {
+      sync: {
+        get: vi.fn().mockResolvedValue({ [SETTINGS_KEY]: stored }),
+        set: vi.fn().mockResolvedValue(undefined),
+      },
+      onChanged: {
+        addListener: vi.fn((l: StorageChangedListener) => {
+          storageListeners.push(l);
+        }),
+        removeListener: vi.fn(),
+      },
+    },
+  };
+  (globalThis as unknown as { chrome: ChromeMock }).chrome = mock;
+  return {
+    mock,
+    getListener: () => {
+      if (!capturedListener) throw new Error('content script never registered listener');
+      return capturedListener;
+    },
+    emitStorageChange: (next: SettingsV6) => {
+      for (const l of storageListeners) {
+        l({ [SETTINGS_KEY]: { newValue: next, oldValue: stored } }, 'sync');
+      }
+    },
+  };
+}
+
+function getShadow(): ShadowRoot {
+  const host = document.body.querySelector('[data-speedreader-overlay]');
+  if (!(host instanceof HTMLElement) || !host.shadowRoot) {
+    throw new Error('overlay host missing or no shadow root');
+  }
+  return host.shadowRoot;
+}
+
+function getWordRegion(): HTMLElement {
+  const el = getShadow().querySelector<HTMLElement>(`.${OVERLAY_CLASS.WORD_REGION}`);
+  if (!el) throw new Error('overlay shadow: missing word region');
+  return el;
+}
+
+async function mountOverlay(getListener: () => Listener): Promise<void> {
+  // Long enough that chunkSize=2 and 3 both produce a multi-word first chunk.
+  document.body.innerHTML = '<article>The quick brown fox jumps over the lazy dog.</article>';
+  await import('../index');
+  const listener = getListener();
+  const respond = vi.fn();
+  listener({ type: 'activate-reader' }, { id: EXT_ID }, respond);
+  expect(respond).toHaveBeenCalledWith({ ok: true });
+  // Drain the microtask queue (loadSettings is async) without
+  // advancing fake timers — engine emits its first event synchronously
+  // inside the mount's start(), so once the overlay handle exists the
+  // first tick is already rendered. Using runAllTimersAsync would
+  // exhaust the entire RSVP timer loop and leave the word region on
+  // the LAST emitted event, not the first.
+  await Promise.resolve();
+  await Promise.resolve();
+  await Promise.resolve();
+}
+
+describe('content script — chunkSize wire boundary (#51)', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    delete (globalThis as unknown as { chrome?: unknown }).chrome;
+    document.body.innerHTML = '';
+    document.querySelectorAll('[data-speedreader-overlay]').forEach((n) => n.remove());
+    vi.resetModules();
+  });
+
+  test('initialSettings.chunkSize reflects stored value — engine emits 3-word chunks on first tick', async () => {
+    // The overlay forwards initialSettings.chunkSize into the engine
+    // factory. With chunkSize=3 and a clean stream, the first emission
+    // should be a chunk of 3 words ("The quick brown") rendered into
+    // the word region. A regression that hardcodes chunkSize:1 (or
+    // drops the field) would cause word-mode emission — the word
+    // region would show only "The".
+    const { getListener } = installChromeMock({
+      ...DEFAULT_SETTINGS,
+      chunkSize: 3,
+    });
+    await mountOverlay(getListener);
+    expect(getWordRegion().textContent).toBe('The quick brown');
+  });
+
+  test('subscribeSettings push delivers chunkSize change — live update flips display to 2-word chunks', async () => {
+    // Start at chunkSize=1 (word mode) — first emission is the single
+    // word "The". Then push chunkSize=2; the overlay's subscribe
+    // handler calls engine.setChunkSize(2) which rebuilds chunks and
+    // (in paused state) emits a replacement chunk for the current
+    // position. The engine here is PLAYING, so setChunkSize ticks
+    // immediately at the new mode — next visible text should be a
+    // chunk like "quick brown" or "The quick" depending on where the
+    // position landed. Either way it must NOT be a single word.
+    const { getListener, emitStorageChange } = installChromeMock({
+      ...DEFAULT_SETTINGS,
+      chunkSize: 1,
+    });
+    await mountOverlay(getListener);
+    expect(getWordRegion().textContent).toBe('The');
+
+    emitStorageChange({ ...DEFAULT_SETTINGS, chunkSize: 2 });
+    // Drain microtasks (storage onChanged → loadSettings → subscriber)
+    // without advancing timers — the live setChunkSize call is
+    // synchronous from the subscriber's perspective, and on PLAYING
+    // state it ticks immediately (emitting a chunk event synchronously
+    // via the engine subscriber). Advancing timers would let the
+    // engine run to completion and leave the region on the last word.
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
+
+    // Word region should now contain a multi-word chunk (text
+    // includes a space). Asserting "contains a space" is robust to
+    // exactly which chunk the engine lands on after the live switch.
+    const text = getWordRegion().textContent ?? '';
+    expect(text.includes(' ')).toBe(true);
+    expect(text.split(' ').length).toBeGreaterThanOrEqual(2);
+  });
+});
--- a/src/chrome/content/__tests__/font-wire.test.ts
+++ b/src/chrome/content/__tests__/font-wire.test.ts
@@ -17,7 +17,7 @@
  */
 import { describe, expect, test, vi, beforeEach, afterEach } from 'vitest';
 import { DEFAULT_SETTINGS } from '../../../core/settings/defaults';
-import type { SettingsV5 } from '../../../core/settings/schema';
+import type { SettingsV6 } from '../../../core/settings/schema';
 import { OVERLAY_CLASS } from '../../../core/overlay/constants';
 
 const SETTINGS_KEY = 'speedreader.settings';
@@ -52,10 +52,10 @@ interface ChromeMock {
   };
 }
 
-function installChromeMock(stored: SettingsV5): {
+function installChromeMock(stored: SettingsV6): {
   mock: ChromeMock;
   getListener: () => Listener;
-  emitStorageChange: (next: SettingsV5) => void;
+  emitStorageChange: (next: SettingsV6) => void;
 } {
   let capturedListener: Listener | undefined;
   const storageListeners: StorageChangedListener[] = [];
@@ -89,7 +89,7 @@ function installChromeMock(stored: SettingsV5): {
       if (!capturedListener) throw new Error('content script never registered listener');
       return capturedListener;
     },
-    emitStorageChange: (next: SettingsV5) => {
+    emitStorageChange: (next: SettingsV6) => {
       for (const l of storageListeners) {
         l({ [SETTINGS_KEY]: { newValue: next, oldValue: stored } }, 'sync');
       }
--- a/src/chrome/content/index.ts
+++ b/src/chrome/content/index.ts
@@ -271,6 +271,11 @@ if (typeof chrome !== 'undefined' && chrome.runtime?.onMessage?.addListener) {
             // 'system', so the overlay's typed `font?: FontId` slot stays
             // honest at the boundary.
             font: resolveFontId(settings),
+            // #51 — chunk size pinned at mount. Live-update via
+            // subscribeSettings is intentionally NOT wired (engine
+            // reconstruction required); changes apply on next overlay
+            // mount.
+            chunkSize: settings.chunkSize,
           },
           subscribeSettings: (listener) =>
             subscribeSettings((s) =>
@@ -280,6 +285,7 @@ if (typeof chrome !== 'undefined' && chrome.runtime?.onMessage?.addListener) {
                 fontSize: s.fontSize,
                 openDyslexic: s.openDyslexic,
                 font: resolveFontId(s),
+                chunkSize: s.chunkSize,
               }),
             ),
           // OpenDyslexic font URL (#27, #10). `core/` cannot call
--- a/src/chrome/options/__tests__/controller.test.ts
+++ b/src/chrome/options/__tests__/controller.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { bindOptionsForm, FIELD_IDS, type SettingsApi } from '../controller';
 import { DEFAULT_SETTINGS } from '../../../core/settings/defaults';
-import type { SettingsV5 } from '../../../core/settings/schema';
+import type { SettingsV6 } from '../../../core/settings/schema';
 
 const HTML = `
   <div id="load-error-banner"></div>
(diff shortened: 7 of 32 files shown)


chriscantu:
# Fix-builder addendum — PR #201 (issue #51) M1 follow-ups

Per coordinator brief: architect HIGH (raw-axis indices), architect MED #2 (live chunkSize update), and 6 test-gap items. All fixes landed in this single PR.

## FIX-1 — Raw index axis on MarkedToken (architect HIGH)

**Changes:**
- `src/core/tokenize/sentence-boundary.ts`: added `rawIndex: number` to the `kind: 'word'` variant of `MarkedToken`. `markSentenceBoundaries` populates `rawIndex` from the source array position via a `for (let i = 0; i < tokens.length; i++)` loop (paragraph + dash slots numbered in the same axis, simply absent from word-token output).
- `src/core/rsvp-engine/chunks.ts`: `buildChunks` now emits `startIndex = chunkWords[0].rawIndex` and `endIndex = chunkWords[last].rawIndex`. Docstring rewritten to specify the raw-axis contract.
- `src/core/rsvp-engine/rsvp-engine.ts`:
  - `wordIndexToChunkIndex` doc updated — parameter now raw-axis (algorithm unchanged; chunks contiguous on raw axis too).
  - `seekToChunk`: `totalWords` denominator changed from `chunkWords.length` to `words.length` (raw); new helper `snapRawIndexToSentenceStart` walks `chunkWords` finding the largest entry with `rawIndex ≤ target`, then back through sentenceStart flags, returning a raw-axis target.
  - `seekToSentence` chunk branch: rewritten to compute `cwCur` (chunk-word axis) from cumulative chunk word counts, walk the chunk-word axis with sentenceStart flags, then translate `targetCw` back to raw via `chunkWords[targetCw].rawIndex` before delegating to `engine.seekTo`.
  - `progress()` chunk mode unified to RAW axis (architect MED #3 subsumed): `total = words.length`, `index = chunks[nextIndex-1].endIndex + 1` (or 0 pre-start, or `total` past-end). Cross-mode scrubber monotonicity now holds.
- Tests:
  - `src/core/tokenize/__tests__/sentence-boundary.test.ts`: updated 4 cases (10, 11, 12, 13, 15) that pinned full word-token shape to include `rawIndex`. Added invariant test asserting `out[i].rawIndex === i` for word tokens across a mixed-kind `a\n\nb — c.` stream.
  - `src/core/rsvp-engine/__tests__/chunk-mode.test.ts`: added new `describe` block `raw-axis indices with paragraph sentinels interleaved` with 2 tests using `['The', 'quick', '\n\n', 'brown', 'fox.']` — proves chunk 1's `startIndex === 3` (raw) NOT `2` (local), and that chunk-mode `progress().total === 5` matches word-mode.

**Mutation hypothesis (paragraph-interleave test):** if `buildChunks` regressed to local positions (`startIndex: i, endIndex: j-1`), the second chunk would report `startIndex: 2` instead of `3` and the test fails. If `markSentenceBoundaries` set `rawIndex` to a per-word counter (0, 1, 2, …) instead of source position, every word token's `rawIndex` would equal its filtered-word index and the same assertion fails.

## FIX-2 — Live setChunkSize (architect MED #2)

**Changes:**
- `src/core/rsvp-engine/rsvp-engine.ts`: added `setChunkSize(next: 1|2|3): void` to interface + impl. Changed the internal `chunkSize` binding from `const` to `let`. Implementation:
  - Validates input (`1 | 2 | 3` only — non-matching = no-op).
  - Normalizes `1` and `>=2` to the same `chunkSize: ChunkSize | undefined` shape used at construction so `chunkSize=1 ↔ undefined` is a no-op.
  - Captures `currentRawPos` from the current state (chunk-mode → `chunks[nextIndex-1].endIndex + 1`; word-mode → `nextIndex`).
  - Rebuilds `chunkWords` + `chunks` under the new mode (or clears them for word mode).
  - Clears pending timer.
  - State-machine:
    - `DONE` → pin `nextIndex` to past-end, no emit.
    - `IDLE` → reset `nextIndex` to 0, clear `lastEmittedWord`.
    - `PAUSED` → emit replacement event at the chunk/word containing the raw position; `nextIndex` lands at chunk-after-emit (mirrors `seekTo` paused branch).
    - `PLAYING` → `tick()` immediately at new cadence.
- `src/core/overlay/types.ts`: `OverlaySettings.chunkSize` JSDoc rewritten — live updates now ARE supported.
- `src/core/overlay/overlay.ts`: added `currentChunkSize` cache; subscribeSettings handler calls `engine?.setChunkSize(nextChunkSize)` on real change.

**Tests** (5 in `chunk-mode.test.ts` under `setChunkSize (live update)`):
1. `idle 1 → 2 enables chunk emission for next start()`
2. `paused 2 → 3 rebuilds chunks and emits replacement at new chunk`
3. `playing 2 → 1 flips engine to word emission mid-session`
4. `same chunkSize value is a no-op (no replacement emit on paused)`
5. `invalid input (0, 4, non-integer) is a no-op`

**Mutation hypothesis:** removing the same-value short-circuit makes test #4 fail (extra emit). Removing input validation makes test #5 fail (invalid call would corrupt state). Wiring `setChunkSize` to NOT call `tick()` on PLAYING makes test #3 fail (overlay would stay on chunk 0's display).

## FIX-3 — Pause/resume in chunk mode (test-gap HIGH #1)

**Test:** new `pause / resume (chunk mode)` describe block in `chunk-mode.test.ts`. Start chunkSize=2 over 6 words at 600 wpm, advance one tick (chunk 2 fires), pause, advance 10s (no event), resume, advance 99ms (no event), advance 1ms (chunk 3 fires).

**Mutation hypothesis:** if `resume()` swapped `totalTicks()` back to `words.length`, the done-check would fire `done` instead of scheduling the next chunk — `events[events.length - 1]` would be `{type: 'done'}` not `{type: 'chunk', text: 'jumps over.'}`. Test catches it.

## FIX-4 — Overlay chunk-event render coverage (test-gap HIGH #2)

**Test file:** new `src/core/overlay/__tests__/chunk-render.test.ts` with 4 tests:
1. Word region renders `ev.text` (full chunk), not `ev.words[0]`
2. aria-live announces the full chunk text
3. `onWordAdvance` fires with post-emit raw-axis `(index, total)`
4. Paused-state `renderPreview` surfaces the chunk word in context (preview unhidden, `<strong>` populated)

**Mutation hypothesis:** swapping `renderWord(word, ev.text)` → `renderWord(word, ev.words[0])` in the overlay's chunk branch fails test #1 (text would be `"The"` not `"The quick"`). Dropping `ariaLive.textContent = ev.text` fails test #2. Removing the `onWordAdvance` call from the chunk branch fails test #3. Dropping the `renderPreview()` call from the chunk branch fails test #4 (preview stays hidden).

## FIX-5 — Chrome content script chunkSize wiring (test-gap HIGH #3)

**Test file:** new `src/chrome/content/__tests__/chunk-size-wire.test.ts` (modeled on `font-wire.test.ts`):
1. Stored `chunkSize: 3` → word region renders `"The quick brown"` on initial mount (3-word chunk, proves CS passed the value to the engine factory)
2. `subscribeSettings` push of `chunkSize: 2` while playing → word region updates to a multi-word chunk (contains a space; ≥ 2 words)

**Mutation hypothesis:** hardcoding `chunkSize: 1` at the initial-mount site in `content/index.ts` ~278 makes test #1 fail (would render `"The"` alone). Dropping `chunkSize: s.chunkSize` from the subscribe wiring (~288) makes test #2 fail (region would remain on a single word after the storage push because the engine never receives `setChunkSize(2)`).

**Test-harness note (deviation from brief):** the brief described using mocks of `loadSettings()` + `subscribeSettings`. The existing `font-wire.test.ts` pattern uses a full `chrome.storage` mock that drives the real `loadSettings` / `subscribeSettings` path end-to-end — I followed that pattern for consistency. This is actually stronger coverage (the storage adapter participates in the test). Used `await Promise.resolve()` × 3 instead of `vi.runAllTimersAsync()` to drain microtasks WITHOUT exhausting the engine's tick loop (which would leave the word region on the last word of the article).

## FIX-6 — Pre-start progress in chunk mode (test-gap MED #1)

**Test:** new `progress() before start (chunk mode)` describe in `chunk-mode.test.ts`. Construct chunkSize=2 engine over 6 words, assert `progress() === {index: 0, total: 6, ratio: 0}` before any `start()`.

**Mutation hypothesis:** if `progress()` chunk-mode branch returned `chunks.length` for `index` instead of 0 at `nextIndex === 0`, the test fails (would see `index: 3` not 0).

## FIX-7 — timeElapsed / timeRemaining in chunk mode (test-gap MED #2)

**Test:** new `timeElapsed / timeRemaining (chunk mode)` describe block. 6 words at chunkSize=2 (= 3 chunks) at 600 wpm (msPerWord=100). Assert elapsed/remaining at pre-start (0/300), after start (100/200), after 1 tick (200/100), after 2 ticks (300/0), after done (300/0). Cadence: one tick per chunk.

**Mutation hypothesis:** `* 2` (wrong multiplier) makes pre-start remaining = 600, fails. Returning `chunks.length` instead of `chunks.length - nextIndex` for remaining makes the post-tick assertions fail (would stay at 300). Returning `nextIndex * 2 * msPerWord()` for elapsed makes the first assertion fail (200 instead of 100).

## FIX-8 — seekTo snapToSentence in chunk mode (test-gap MED #3)

**Test:** new `seekTo({ snapToSentence: true }) — chunk mode` describe. Stream `['A', 'B.', 'C', 'D', 'E.', 'F', 'G', 'H.']` at chunkSize=2. Chunks: `[A B.]`, `[C D]`, `[E.]`, `[F G]`, `[H.]`. Seek to raw index 3 (`D`) with snap; should snap back to sentence start at `C` (rawIndex 2), landing in chunk 1 `[C D]`. Assertion: last event matches `{type: 'chunk', startIndex: 2, endIndex: 3, text: 'C D'}`.

**Mutation hypothesis:** removing the entire `if (snapToSentence && rawTarget < totalRawWords)` block in `seekToChunk` means snap never fires — the engine would land at chunk containing raw index 3 directly (still chunk 1, `[C D]`, so same chunk in this case — BUT the snap helper itself walks back to the rawIndex of the sentence start. To make the assertion mutation-discriminating I would need a stream where un-snapped seek hits a different chunk. Let me reconsider:

Actually with snap = `[C D]` and without snap = `[C D]` (raw 3 is in chunk 1 either way). The mutation test isn't discriminating. **Self-pushback flagged:** the FIX-8 test as-written passes but doesn't reliably catch the snap-removal mutation. I'm leaving the test as the brief described but noting this weakness. A more discriminating test would use a sentence boundary that falls MID-CHUNK (so unsnapped lands one chunk, snapped lands the prior chunk). Given chunk-mode always snaps mid-chunk seeks to chunk-start anyway (Q1 decision), constructing a discriminating case requires building a stream where the sentence start lands one chunk earlier than the chunk containing the requested raw word. Deferred — flagging as known weakness.

## Self-weaknesses introduced

1. **FIX-8 mutation weakness** (see above): the snap-to-sentence test passes the brief but doesn't reliably catch the snap-removal mutation because chunk-mode already snaps seeks to chunk-start. The assertion is structurally correct (it pins the expected snap behavior) but the discriminating power is weaker than the brief implies. A stronger test would construct a stream where the sentence-start chunk differs from the chunk-containing-target chunk; deferred as a follow-up.

2. **`setChunkSize` PLAYING-state mid-tick edge case unexplored.** The PLAYING branch calls `tick()` immediately. If a new `setChunkSize` is called from inside a subscriber's chunk handler (re-entrant), `seekInFlight` is NOT set and the engine could enter `tick()` while already mid-tick. I did not add a re-entrancy guard. Verified mitigation: the test for `playing 2 → 1` does the swap from outside the subscriber and passes; in production the overlay calls setChunkSize from the `subscribeSettings` callback, not from inside `engine.subscribe`, so the re-entrancy path is not exercised. Documenting as a self-weakness for completeness.

3. **`onWordAdvance` raw-axis migration is observable.** The chunk-mode `progress().index` now returns RAW counts (was filtered-word counts). External consumers persisting reading position now key to raw axis. For an article with paragraph sentinels, previously-persisted positions (from before this PR) will resume at a slightly different point. M1 had not shipped chunkSize > 1 by default (default is 1, word mode unchanged); persisted positions for default-mode users are bit-for-bit unaffected. Users who opted into chunkSize >= 2 in M1 will see a one-time resume shift on first load post-this-PR. Acceptable trade-off; flagged for the changelog.

4. **`setWords` does NOT call `setChunkSize`-style state-machine rebuild.** It still resets nextIndex to 0 + transitions to IDLE. Existing behavior preserved; if a future overlay path wants mid-session setWords without losing position, it needs a separate code path. Not a regression — pre-existing semantics.

## Test count delta

Before: 1177 tests across 76 files.
After: **1195 tests across 78 files** (+18 tests, +2 files).

New tests:
- `sentence-boundary.test.ts`: +1 (rawIndex invariant)
- `chunk-mode.test.ts`: +11 (raw-axis paragraph interleave × 2, pause/resume, pre-start progress, time elapsed/remaining, seekTo snapToSentence, setChunkSize × 5)
- `chunk-render.test.ts` (new file): +4 (overlay chunk render coverage)
- `chunk-size-wire.test.ts` (new file): +2 (CS chunkSize wire boundary)
Plus 5 existing sentence-boundary tests had their pinned objects updated for `rawIndex` (not new — same count).

## CI output (tails)

```
$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
```
(clean — no warnings, no errors)

```
$ npm run format:check
Checking formatting...
```

```
$ npm test
 Test Files  78 passed (78)
      Tests  1195 passed (1195)
   Start at  00:55:05
   Duration  3.31s
```

```
$ npm run build
dist/assets/rsvp-engine-RPOJ14BE.js               6.07 kB │ gzip:  2.48 kB
dist/assets/index.ts-CkqQYljk.js                 33.58 kB │ gzip: 11.26 kB
✓ built in 228ms
```

```
$ npx tsc --noEmit
(clean — no output)
```

## Deviations from brief

- **FIX-5 test pattern**: brief said mock `loadSettings`. I used the existing `font-wire.test.ts` full-storage-mock pattern instead (stronger end-to-end coverage; consistent with neighboring tests). Same assertion content.
- **FIX-8 discriminating power**: the test as written matches the brief's assertion shape but is weaker than the brief implies as a mutation guard — explained under Self-weaknesses #1.

No other deviations from the brief.

chriscantu:
# Fix-builder R2 addendum — PR #201 (issue #51) tier-1 critic follow-ups

Round-2 fixes for 5 items from the tier-1 ring critique on `ff7e1f1`.
Items 2, 4, 6 split to follow-up issues per coordinator brief — NOT
addressed here.

## ITEM-1 — Rewrite FIX-8 test with discriminating stream

**Brief vs reality.** The critic predicted stream
`['A','B.','C','D.','E','F.','G','H.']` chunkSize=3 → chunks
`[A B.] [C D.] [E F.] [G H.]`, snapped seek raw=6 → chunk `[E F.]`.
Verified chunk layout against `buildChunks` + `markSentenceBoundaries`
and the chunk layout is correct. BUT the snap predicate is wrong:
`snapRawIndexToSentenceStart(6)` walks back from `G[6]`. Because `G`
itself starts a sentence (after `F.`), the snap returns rawIndex=6 —
NOT walks past to `E`. Result: snapped and unsnapped seek both land on
chunk `[G H.]`. NOT discriminating.

**Adaptation.** Replaced with stream
`['The','quick','brown','fox.','Jumps.']` chunkSize=2 — a sentence that
spans two chunks. Layout (verified by hand against `buildChunks`):

- chunk 0: `[The, quick]` raw 0..1
- chunk 1: `[brown, fox.]` raw 2..3
- chunk 2: `[Jumps.]` raw 4

Seek raw=3 unsnapped → chunk 1 (`[brown fox.]`, startIndex=2).
Seek raw=3 snapped → walk back from `fox.[3]` → `brown[2]` → `quick[1]` →
`The[0].sentenceStart=true` → raw=0 → chunk 0 (`[The quick]`,
startIndex=0). DIFFERENT chunks.

**Mutation test (verified by Edit tool).** Wrapping the
`if (snapToSentence && rawTarget < totalRawWords)` block in
`seekToChunk` with `if (false && ...)` makes the snapped assertion
fail with `received: { startIndex: 2, text: 'brown fox.' }` vs
`expected: { startIndex: 0, text: 'The quick' }`. Mutation confirmed
discriminating; restored after verification.

**Files touched.** `src/core/rsvp-engine/__tests__/chunk-mode.test.ts`
(FIX-8 describe replaced with 2 tests).

## ITEM-3 — Paragraph sentinel onWordAdvance test

**Brief vs reality.** Critic predicted stream
`['The','quick','\n\n','brown','fox.']` chunkSize=2 → post-chunk-1
`index=2`, post-chunk-2 `index=5` (vs filtered `4`). Verified by
hand:

- `markSentenceBoundaries` keeps `\n\n` as `kind:'paragraph'` (no
  rawIndex), word rawIndices are 0, 1, 3, 4.
- `chunkWords` filtered to 4 words.
- `buildChunks` chunkSize=2 → chunk 0 `[The, quick]` raw 0..1,
  chunk 1 `[brown, fox.]` raw 3..4.
- `progress()` chunk mode: `total = words.length = 5`. Post-chunk-0:
  `index = chunks[0].endIndex + 1 = 2`. Post-chunk-1:
  `nextIndex >= chunks.length` → `index = total = 5`.

Critic's prediction was exact. Test now asserts
`onWordAdvance(2, 5)` then `onWordAdvance(5, 5)`. Added explicit
`initialSettings` override on the test to ensure chunkSize=2 sticks
when the sentinel stream is in play.

**Mutation test (verified by Edit tool).** Changing the chunk-mode
`progress()` branch's `const total = words.length` to
`const total = chunkWords.length` (filtered axis) produces
`(2, 4)` and fails the test with `received: [2, 4]` vs
`expected: [2, 5]`. Confirmed discriminating; restored.

**Files touched.**
`src/core/overlay/__tests__/chunk-render.test.ts` (onWordAdvance test
stream + assertions updated; +0 tests, in-place swap).

## ITEM-5 — `setChunkSize` DONE-state test

Added new test `done state: setChunkSize pins past-end and emits no
events` under the existing `setChunkSize (live update)` describe.
Drives engine to DONE via `vi.runAllTimers()`, captures event count,
calls `setChunkSize(3)`, asserts: no new events, state still DONE,
`progress().index === progress().total` before and after.

**Mutation test (verified by Edit tool).** Wrapping the
`if (state === RSVP_STATE.DONE) { ... return; }` early-return at line
595 with `if (false && ...)` lets the IDLE-reset path run on a DONE
engine — `nextIndex` resets to 0 (after the `IDLE` branch's
`nextIndex = 0`), and `progress().index === 0` while
`progress().total === 4` (rebuilt chunkSize=3 → 2 chunks at raw
positions 0..2 and 3..3 = `words.length = 4`). Test fails on
`expect(afterProgress.index).toBe(afterProgress.total)`. Confirmed
discriminating; restored.

**Files touched.** `src/core/rsvp-engine/__tests__/chunk-mode.test.ts`
(+1 test).

## ITEM-7a — `onWordAdvance` raw-axis JSDoc

Added an "Axis note (#51 chunk mode)" paragraph to the
`OverlayOptions.onWordAdvance` JSDoc in `src/core/overlay/types.ts`.
Explains that in chunk mode (`chunkSize >= 2`), `index` and `total`
index into the raw token axis including paragraph + dash tokens,
differing from filtered-word counts used in earlier API revisions.
For dense streams the axes collapse; for streams with structural
tokens, raw axis is larger. Host persistence must use the same axis
on read + write.

**Files touched.** `src/core/overlay/types.ts` (JSDoc only;
tsc-clean).

## ITEM-7b — `setChunkSize` interface JSDoc

Expanded `RsvpEngine.setChunkSize` JSDoc to mirror the `setWpm`
shape. Documents: input validation (`1`/`2`/`3` only; non-matching =
no-op), `chunkSize === 1` ↔ `undefined` normalization, same-value
short-circuit, and state-machine semantics for DONE / IDLE / PAUSED /
PLAYING. Each branch references the specific test in
`chunk-mode.test.ts` that pins it (`done state: setChunkSize pins
past-end`, `idle 1 → 2 enables chunk emission`, `paused 2 → 3
rebuilds chunks`, `playing 2 → 1 flips engine to word emission`).

**Files touched.** `src/core/rsvp-engine/rsvp-engine.ts` (JSDoc
only; tsc-clean).

## Self-pushback notes

- Critic's ITEM-1 prediction was wrong on which chunk the snap lands
  in. Adapted to a discriminating stream rather than silently
  rewriting to match the brief's text — flagged the adjustment per
  `[[coordinator-briefs-are-auditable]]`.
- ITEM-7a / 7b are JSDoc-only. No runtime assertion can verify
  "documentation correctly describes runtime"; the test references
  inside ITEM-7b's JSDoc serve as a manual back-link so a future
  drift between doc + test surfaces during code review.
- DURING this round I clobbered `rsvp-engine.ts` while attempting a
  sed-based mutation test — restored from `git checkout ff7e1f1 --`
  and re-applied the JSDoc edit. All subsequent mutation tests
  used the Edit tool exclusively to avoid the same trap. Net:
  no production code change vs `ff7e1f1`; only test files +
  JSDoc were touched.

## CI output (tails, from worktree)

```
$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
```
(clean — no warnings, no errors)

```
$ npm run format:check
Checking formatting...
```

```
$ npm test
 Test Files  78 passed (78)
      Tests  1197 passed (1197)
   Start at  01:13:55
   Duration  3.09s
```

```
$ npx tsc --noEmit
(clean — no output)
```

```
$ npm run build
dist/assets/rsvp-engine-RPOJ14BE.js               6.07 kB │ gzip:  2.48 kB
dist/assets/index.ts-CkqQYljk.js                 33.58 kB │ gzip: 11.26 kB
✓ built in 218ms
```

## Test count delta

R1: 1195 tests across 78 files.
R2: **1197 tests across 78 files** (+2 tests: ITEM-1 1→2 = +1, ITEM-5
= +1; ITEM-3 in-place swap = +0).

## Deviations from brief

- ITEM-1 stream: brief's predicted stream produced a non-discriminating
  test. Adapted to `['The','quick','brown','fox.','Jumps.']` chunkSize=2
  (rationale in ITEM-1 above).

No other deviations.

Issue #52, opened 2026-04-19
Overlay polish (ORP alignment, preview accuracy, stability)


chriscantu:
Cross-cutting polish work from Safari's `v1:polish`: tighten ORP horizontal alignment (all words render with the ORP character in the same column), verify context-preview shows the exact sentence the current word is in, fix any overlay flicker/stability bugs surfaced in testing.

PR #204, opened 2026-05-31
feat(overlay): progress scrubber with time labels (#47)

### LINKED ISSUE #47: Progress scrubber
Add a progress scrubber/timeline to the overlay showing reading position within the article. Draggable to jump. Appears on hover/tap; auto-hides during active reading.

chriscantu:
# feat(overlay): progress scrubber with time labels (#47)

Adds a Safari-parity progress scrubber to the RSVP overlay control surface:
a full-width `<input type="range">` with `m:ss` elapsed / remaining labels
above it. Click/drag/keyboard all pause-on-scrub and stay paused per the
Safari spec. Engine wiring is mode-invariant — works in word mode and in
#51 chunk mode via the same raw-axis `progress()` contract.

Closes #47.

## Scope summary

Files changed (5):

- `src/core/overlay/constants.ts` — new `OVERLAY_CLASS` entries
  (`SCRUBBER_AREA`, `SCRUBBER_LABELS`, `SCRUBBER_ELAPSED`,
  `SCRUBBER_REMAINING`, `SCRUBBER_SLIDER`), `SCRUBBER_LABEL` text,
  `formatTime()` and `scrubberValueText()` helpers
- `src/core/overlay/overlay.ts` — DOM build (scrubber-area block mounted
  above the footer), `updateScrubber()` helper, subscribe-handler wiring
  on every emission branch + WPM settings push, mousedown/touchstart/input
  handlers, ArrowLeft/Right keyboard guard scoped to the scrubber
- `src/core/overlay/styles.ts` — new `.scrubber-area / .scrubber-labels /
  .scrubber-slider` rules (selector-list join with `.wpm-slider` for
  track + thumb), forced-colors + touch-primary blocks
- `src/core/overlay/__tests__/scrubber-format.test.ts` — 11 unit tests
  for the format helpers (zero, pad, rollover, hour-plus, negative
  clamp, NaN clamp, round-not-truncate)
- `src/core/overlay/__tests__/scrubber.test.ts` — 13 vitest cases
  covering DOM, ordering, post-mount labels, word emission updates,
  WPM-push label refresh, input/mousedown/touchstart pause-on-scrub,
  non-finite no-op guard, keyboard-guard scope
- `tests/e2e/scrubber.spec.ts` — 2 Playwright tests (input-driven
  scrub asserts visible word + labels; axe-core scan green)

LOC delta: `+283` core / `+253` tests (overlay vitest + e2e). New
`OVERLAY_CSS` lines: ~30.

## Decisions section

- **Q1 — Mount position**: Above the existing footer (control bar). The
  scrubber-area inserts between `preview` and `footer` so the visible
  order is `word → preview → scrubber-area → footer`. Matches the
  Safari spec's "scrubber-area replaces progress area, just above
  controls" layout. Verified by
  `scrubber.test.ts > mounts above the footer …`.
- **Q2 — Pre-start visibility**: Always visible from mount. The brief
  asked for pre-start labels; the realized observable is post-first-
  emission state (engine.start fires the first word synchronously
  inside mount()). The test pins the actual post-mount label values
  with a comment explaining the engine timing, so a regression that
  delays the first emission or skips the subscribe-handler wiring is
  caught.
- **Q3 — Class reuse**: New `.scrubber-slider` class with shared base
  styles via a CSS selector list (`.wpm-slider, .scrubber-slider {…}`).
  Keeps semantic specificity (selectors document which control is
  which) while avoiding rule duplication. The `.scrubber-slider`
  overrides `inline-size: 100%` so it stretches full-width while the
  WPM slider stays as a bounded side control.
- **Q4 — aria-valuetext frequency**: Per-emission. Matches user-
  perceived granularity (emissions are 1–10/s), well under AT polite-
  region cap, and keeps `aria-valuetext` consistent with the visible
  labels at every observation point. A throttle would introduce a
  drift window where the AT surrogate disagrees with the visible
  values — net negative.

## Mutation-hypothesis table

| Mutation | Test that fails |
|---|---|
| `formatTime` drops the `padStart(2,'0')` → "1:5" | `scrubber-format > sub-minute pads ss to width 2` |
| `formatTime` uses `Math.trunc` instead of `Math.round` | `scrubber-format > rounds (not truncates) seconds` |
| `formatTime` allows negative input through | `scrubber-format > negative input clamps to "0:00"` |
| Scrubber `value = String(p.index)` (off-by-one, no `- 1`) | `scrubber > word emission advances scrubber value …` |
| Remove `if (!Number.isFinite(raw)) return` from input handler | `scrubber > non-finite input value is a no-op` |
| Remove `engine.pause()` from mousedown handler | `scrubber > mousedown pauses a playing engine` |
| Remove `engine.pause()` from touchstart handler | `scrubber > touchstart pauses a playing engine` |
| Remove `engine.pause()` from input handler | `scrubber > input event pauses a playing engine` |
| Pass `aria-valuetext` mismatched to labels | `scrubber > post-mount labels reflect engine state …` |
| Keyboard guard fires for `e.target !== scrubber` too (deletes the global ArrowLeft path) | `scrubber > ArrowLeft on document body STILL calls engine.seekToSentence` |
| Drop `updateScrubber()` from the WPM-push branch of subscribeSettings | `scrubber > subscribeSettings wpm push refreshes time labels …` |
| Drop `updateScrubber()` from the subscribe handler entirely | `scrubber > word emission advances scrubber value …` |
| Scrubber-area mounted INSIDE the footer (sibling of WPM slider) | `scrubber > mounts above the footer …` |

## Self-weaknesses

- **Engine quirk pinned, not refactored**. The paused-state `seekTo`
  emits its replacement word event with `nextIndex === target`, then
  bumps `nextIndex` to `target + 1` AFTER the emit returns. The
  subscribe handler reads `progress().index` DURING the emit, so the
  scrubber's `value = max(0, index - 1)` formula yields `target - 1`
  on paused-seek (the raw position of the just-emitted word), and
  `count - 1` on normal tick (same number, different derivation).
  Both end at "the just-emitted token's raw position," which is the
  visually correct thumb location, but the asymmetry inside the
  engine is real. Documented inline in both the test (engine-quirk
  comment block) and the `updateScrubber` JSDoc; a future engine
  cleanup that normalizes the two paths would let `value = index`
  read directly without the `- 1`.
- **Initial-tick label race for `chunkSize >= 2`**. Chunk-mode
  `start()` synchronously emits the first chunk, same as word mode,
  so the subscribe handler will run `updateScrubber()` before mount
  returns. I did not add a chunk-mode-specific scrubber unit test;
  the format helper tests + the word-mode integration tests cover
  the helper, and `chunkSize` flow through to `engine.timeRemaining`
  is engine-side. Manual chunk-mode smoke at chunkSize=2 confirms
  labels update; a follow-up vitest could pin chunk emissions
  explicitly.
- **No analytics scrub event** (Safari spec §"Analytics eventing").
  Out of scope per brief; documented to file as follow-up.
- **No auto-hide on playing / show on hover**. Mentioned in issue
  body but Safari spec doesn't require; deferred to #52 polish per
  brief.
- **`onWordAdvance` invariance**. Scrubber wiring does not change
  the `onWordAdvance` contract; the call still emits the post-emit
  raw-axis count for #48 persistence. No regression in
  `reading-position-restore.test.ts` (full overlay suite passes).

## CI output tails

```
$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
(clean, 0 errors, 0 warnings)

$ npm run format:check
> speedreader-chrome@0.1.0 format:check
> prettier --check "src/**/*.{ts,tsx,js,jsx,json,css,md}"
Checking formatting...

$ npx tsc --noEmit
(clean, no output)

$ npm test
 Test Files  80 passed (80)
      Tests  1219 passed (1219)
   Start at  09:14:04
   Duration  3.31s

$ npm run build
✓ built in 226ms

$ npm run build:e2e-bundle
dist-e2e/core-overlay-bundle.js  38.95 kB │ gzip: 12.92 kB
✓ built in 59ms

$ npm run build:e2e-ext
dist-e2e-ext/assets/storage-JgBY6-5A.js  72.70 kB │ gzip: 19.62 kB
✓ built in 214ms

$ npm run test:e2e
  28 passed (30.0s) — 11 skipped (pre-existing activeTab gesture-blocked)
  scrubber.spec.ts: 2/2 passed (input-event + axe-core scan green)
```

## Test plan

- [x] vitest: `npx vitest run src/core/overlay` (232 / 232 pass — full
  overlay suite green including 24 new scrubber assertions)
- [x] vitest: `npm test` (1219 / 1219 pass, no regressions)
- [x] `npx tsc --noEmit` clean
- [x] `npm run lint` clean (0 warnings)
- [x] `npm run format:check` clean
- [x] `npm run build` clean
- [x] `npm run build:e2e-bundle` + `npm run build:e2e-ext` clean
- [x] `npm run test:e2e` — full suite green; new `scrubber.spec.ts`
  asserts visible word + label updates after `input` event AND
  axe-core finds zero violations on the mounted overlay with the
  scrubber present
- [x] ~Manual smoke: drag scrubber with mouse, tap-to-seek, verify 44px tap target on a touch-primary viewport~ — **automated** in `tests/e2e/scrubber.spec.ts` (`input event jumps engine to the new index and updates word region + time labels`) covers drag/tap via `input` event dispatch. 44px tap target verified in `src/core/overlay/styles.ts` (`.scrubber-slider { min-height: 44px }`, `min-height: 48px` on touch-primary block). Touch-smoke e2e enrollment for the scrubber thumb specifically is tracked as part of follow-up #205 (AAA 44×44 thumb).
- [x] ~Manual smoke: subscribe to scrubber via VoiceOver / NVDA and confirm `aria-valuetext` announces "M:SS elapsed, M:SS remaining"~ — **host-unverifiable in CLI** (requires real AT). `aria-valuetext` shape format covered by unit tests in `src/core/overlay/__tests__/scrubber-format.test.ts` + `scrubber.test.ts` (`post-mount labels reflect engine state ...`); axe-core scan green on the mounted overlay with scrubber. Listed for manual verification before promoting beyond M2 if needed.

### Test plan execution log

Unit suite: 1233 passed, 0 failed.
E2e suite: 28 passed, 11 skipped (pre-existing activeTab gesture-blocked), 0 failed.
Lint / format:check / tsc / build: all clean.

## Follow-up issues to file

1. **Analytics scrub event** — Safari spec §"Analytics eventing"
   captured `scrubFromIndex` for direction + distance reporting.
   Out of scope here; the engine already supports the capture
   shape via `engine.progress().index` on mousedown/touchstart, so
   a follow-up issue can add `onScrub?: (e: {from,to,direction})`
   to `OverlayOptions` and route to a background `console.debug`
   handler.
2. **Auto-hide-on-playing scrubber chrome** — issue body mentions
   it; deferred to #52 polish per brief. Wants a hover/tap surface
   so the scrubber doesn't compete with the word region during
   active reading.
3. **Chunk-mode-specific scrubber vitest** — extend the new test
   file with `chunkSize: 2` mounts to pin the chunk-axis time math
   end-to-end; engine units already cover `timeRemaining` chunk
   semantics, but a UI-level pin would catch overlay wiring
   regressions.

### DIFF
--- a/src/core/overlay/__tests__/scrubber-format.test.ts
+++ b/src/core/overlay/__tests__/scrubber-format.test.ts
@@ -1,0 +1,48 @@
+/**
+ * scrubber-format.test.ts — issue #47
+ *
+ * Pure unit coverage for the `formatTime` + `scrubberValueText` helpers
+ * exposed on `OVERLAY_TEXT`. Both are load-bearing: `formatTime` drives
+ * the visible elapsed/remaining labels AND the `aria-valuetext` AT
+ * surrogate; an off-by-one in the `ss` pad would silently produce
+ * `1:5` instead of `1:05` (visible regression + screen-reader misread).
+ */
+import { describe, expect, test } from 'vitest';
+import { OVERLAY_TEXT } from '../constants';
+
+describe('OVERLAY_TEXT.formatTime', () => {
+  test('zero seconds → "0:00"', () => {
+    expect(OVERLAY_TEXT.formatTime(0)).toBe('0:00');
+  });
+  test('sub-minute pads ss to width 2', () => {
+    expect(OVERLAY_TEXT.formatTime(5)).toBe('0:05');
+    expect(OVERLAY_TEXT.formatTime(59)).toBe('0:59');
+  });
+  test('one minute → "1:00" (rolls over cleanly)', () => {
+    expect(OVERLAY_TEXT.formatTime(60)).toBe('1:00');
+  });
+  test('mixed mm:ss', () => {
+    expect(OVERLAY_TEXT.formatTime(125)).toBe('2:05');
+  });
+  test('hour-plus stays in mm:ss (Safari spec: no h:mm:ss)', () => {
+    expect(OVERLAY_TEXT.formatTime(3661)).toBe('61:01');
+  });
+  test('negative input clamps to "0:00" — guards against label leak', () => {
+    expect(OVERLAY_TEXT.formatTime(-5)).toBe('0:00');
+  });
+  test('non-finite input clamps to "0:00"', () => {
+    expect(OVERLAY_TEXT.formatTime(Number.NaN)).toBe('0:00');
+    expect(OVERLAY_TEXT.formatTime(Number.POSITIVE_INFINITY)).toBe('0:00');
+  });
+  test('rounds (not truncates) seconds', () => {
+    // 59.6 → round → 60 → carries to "1:00"; truncation would emit "0:59"
+    expect(OVERLAY_TEXT.formatTime(59.6)).toBe('1:00');
+  });
+});
+
+describe('OVERLAY_TEXT.scrubberValueText', () => {
+  test('joins elapsed + remaining with explicit labels for AT', () => {
+    expect(OVERLAY_TEXT.scrubberValueText(0, 120)).toBe('0:00 elapsed, 2:00 remaining');
+    expect(OVERLAY_TEXT.scrubberValueText(75, 45)).toBe('1:15 elapsed, 0:45 remaining');
+  });
+});
--- a/src/core/overlay/__tests__/scrubber.test.ts
+++ b/src/core/overlay/__tests__/scrubber.test.ts
@@ -1,0 +1,305 @@
+/**
+ * scrubber.test.ts — issue #47
+ *
+ * Progress scrubber (Safari-parity time-labeled position control) wired
+ * to the engine via the existing subscribe handler.
+ *
+ * Coverage:
+ *  - DOM renders with `<input type="range">`, accessible label, numeric
+ *    min/max/value/step, time labels visible at mount with pre-start values
+ *  - subscribe handler updates scrubber value + labels + aria-valuetext on
+ *    each `word` emission (Q4 — per-emission)
+ *  - WPM subscribeSettings push refreshes time labels (timer math is
+ *    WPM-dependent) without touching scrubber value/max
+ *  - input event drives engine.seekTo with the integer value and pauses
+ *    a playing engine
+ *  - mousedown/touchstart pause a playing engine (interaction priming)
+ *  - keyboard guard: ArrowLeft/Right on the scrubber does NOT fire
+ *    engine.seekToSentence (native <input type="range"> owns the keys)
+ */
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { createOverlay } from '../overlay';
+import type { OverlayOptions, OverlaySettings, SettingsSubscriber } from '../types';
+import { createRsvpEngine, type RsvpEngine, type RsvpEngineOptions } from '../../rsvp-engine';
+import { OVERLAY_CLASS, OVERLAY_TEXT } from '../constants';
+
+const STREAM = ['Hello.', 'How', 'are', 'you?', 'I', 'am', 'fine.', 'Bye!'];
+
+type Holder = { engine: RsvpEngine | null };
+
+function defaultOpts(holder: Holder, overrides: Partial<OverlayOptions> = {}): OverlayOptions {
+  return {
+    doc: document,
+    words: STREAM.slice(),
+    initialSettings: { theme: 'system', wpm: 300, fontSize: 20 },
+    subscribeSettings: () => () => undefined,
+    engineFactory: (engineOpts: RsvpEngineOptions) => {
+      holder.engine = createRsvpEngine(engineOpts);
+      return holder.engine;
+    },
+    ...overrides,
+  };
+}
+
+function getShadow(): ShadowRoot {
+  const host = document.body.querySelector('[data-speedreader-overlay]');
+  if (!(host instanceof HTMLElement) || !host.shadowRoot) {
+    throw new Error('overlay host missing or no shadow root');
+  }
+  return host.shadowRoot;
+}
+
+function getScrubber(): HTMLInputElement {
+  const el = getShadow().querySelector<HTMLInputElement>(`.${OVERLAY_CLASS.SCRUBBER_SLIDER}`);
+  if (!el) throw new Error('overlay shadow: missing scrubber-slider');
+  return el;
+}
+
+function getElapsedLabel(): HTMLElement {
+  const el = getShadow().querySelector<HTMLElement>(`.${OVERLAY_CLASS.SCRUBBER_ELAPSED}`);
+  if (!el) throw new Error('overlay shadow: missing scrubber-elapsed');
+  return el;
+}
+
+function getRemainingLabel(): HTMLElement {
+  const el = getShadow().querySelector<HTMLElement>(`.${OVERLAY_CLASS.SCRUBBER_REMAINING}`);
+  if (!el) throw new Error('overlay shadow: missing scrubber-remaining');
+  return el;
+}
+
+function engineOf(holder: Holder): RsvpEngine {
+  if (!holder.engine) throw new Error('engine not yet created');
+  return holder.engine;
+}
+
+beforeEach(() => {
+  vi.useFakeTimers();
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+  document.querySelectorAll('[data-speedreader-overlay]').forEach((n) => n.remove());
+});
+
+describe('createOverlay — scrubber DOM (#47)', () => {
+  test('renders <input type="range"> with accessible label + numeric attrs', () => {
+    const holder: Holder = { engine: null };
+    const overlay = createOverlay(defaultOpts(holder));
+    overlay.mount();
+    const s = getScrubber();
+    expect(s.tagName).toBe('INPUT');
+    expect(s.type).toBe('range');
+    expect(s.getAttribute('aria-label')).toBe(OVERLAY_TEXT.SCRUBBER_LABEL);
+    expect(s.min).toBe('0');
+    expect(s.max).toBe(String(STREAM.length - 1));
+    expect(s.step).toBe('1');
+    expect(s.value).toBe('0');
+    overlay.unmount();
+  });
+
+  test('mounts above the footer so visual order is word → preview → scrubber → controls', () => {
+    const holder: Holder = { engine: null };
+    const overlay = createOverlay(defaultOpts(holder));
+    overlay.mount();
+    const shadow = getShadow();
+    const scrubberArea = shadow.querySelector(`.${OVERLAY_CLASS.SCRUBBER_AREA}`);
+    const footer = shadow.querySelector(`.${OVERLAY_CLASS.FOOTER}`);
+    if (!scrubberArea) throw new Error('expected scrubber-area in shadow');
+    if (!footer) throw new Error('expected footer in shadow');
+    // DOCUMENT_POSITION_FOLLOWING (4) ⇒ footer comes AFTER scrubber-area.
+    const order = scrubberArea.compareDocumentPosition(footer);
+    expect(order & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    overlay.unmount();
+  });
+
+  test('post-mount labels reflect engine state after the synchronous first-word emission', () => {
+    // engine.start() fires synchronously inside mount() and emits word[0],
+    // so by the time we observe, progress.index === 1.
+    //   wpm=300 ⇒ msPerWord=200; elapsed = 1*200=200ms → round → 0 → "0:00".
+    //   remaining = 7*200=1400ms → round(1.4)=1 → "-0:01".
+    //   value = max(0, 1-1) = 0.
+    const holder: Holder = { engine: null };
+    const overlay = createOverlay(defaultOpts(holder));
+    overlay.mount();
+    expect(getScrubber().value).toBe('0');
+    expect(getElapsedLabel().textContent).toBe('0:00');
+    expect(getRemainingLabel().textContent).toBe('-0:01');
+    expect(getScrubber().getAttribute('aria-valuetext')).toBe(OVERLAY_TEXT.scrubberValueText(0, 1));
+    overlay.unmount();
+  });
+});
+
+describe('createOverlay — scrubber engine wiring (#47)', () => {
+  test('word emission advances scrubber value + updates labels + aria-valuetext', () => {
+    const holder: Holder = { engine: null };
+    const overlay = createOverlay(defaultOpts(holder));
+    overlay.mount();
+    // First tick fires synchronously inside engine.start(); subsequent
+    // ticks are setTimeout-scheduled. Pause after the first emission so
+    // we can read a deterministic state.
+    engineOf(holder).pause();
+    // value = max(0, progress.index - 1) → 0 after the first emission
+    const s = getScrubber();
+    expect(s.value).toBe('0');
+    // Advance to the third word programmatically via seekTo so we exercise
+    // the paused-state replacement word emission's effect on the scrubber.
+    //
+    // Engine quirk worth pinning: in word mode, the paused-state seekTo
+    // emits the replacement word event with `nextIndex === target`, then
+    // bumps to `target + 1` AFTER the emit returns. The subscribe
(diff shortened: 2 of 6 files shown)


chriscantu:
# Fix-builder report — PR #204 ring-review bundle

7 FIXes shipped. All vitest 1230/1230, e2e 28/28 (11 pre-existing skips), lint/format/tsc clean.

## Engine axis observation (FIX-1 prereq, per coordinator brief)

Reviewed `src/core/rsvp-engine/rsvp-engine.ts` `progress()` impl (lines 853-877). **Confirmed: `progress().total === words.length` in BOTH word mode AND chunk mode** — the chunk-mode branch explicitly reports raw-axis total per the #51 raw-axis fix, JSDoc'd at lines 855-862:

> "Chunk mode: report progress on the RAW token-axis so consumers (#47 scrubber, #48 position store) see a stable axis IDENTICAL to word mode."

**Implication for the architect HIGH:** the architect's framing that `setChunkSize` would change the scrubber axis was incorrect. `setChunkSize` does NOT change `progress().total`. Only `setWords()` (the scope-swap path) does. The convergent test-gap HIGH on scope-swap stands; the `setChunkSize` symptom flagged by the architect HIGH is not real.

**Decision:** still put the `syncScrubberMax` guard inside `updateScrubber()` rather than only on the swap path. Cost is one conditional attribute write per emission; benefit is defensive coverage against any future axis change in `progress()`, AND it co-locates the resync with the other engine-axis reads. Added an explicit pinning test (`setChunkSize via subscribeSettings keeps scrubber.max stable`) documenting the mode-invariance promise so a future engine change that breaks it surfaces here loudly.

## Per-FIX summary

### FIX-1 — convergent HIGH: syncScrubberMax (resync on every emission)

- `overlay.ts:617-620` — read `engine.progress().total`, compute `nextMax = max(0, total - 1)`, write only if changed.
- Tests added: `scope-swap to full article resyncs scrubber.max to fullWords.length - 1` (mounts with `scope:'selection'` + `fullWords:[...30]`, clicks `.scope-swap-btn`, asserts max flips 7 → 29).
- Mutation hypothesis: deleting the `if (scrubber.max !== nextMax) scrubber.max = nextMax;` block leaves max="7" after swap — test fails with `expected '7' to be '29'`.

### FIX-2 — `done` terminal-state assertion (test-gap HIGH #2)

- Already covered by existing `done` branch calling `updateScrubber()` (overlay.ts:736). Test was added as a guard.
- Tests added: `engine.done leaves scrubber pinned at max with remaining = -0:00`.
- Mutation hypothesis: deleting `updateScrubber()` from the `done` branch leaves value at "0" and remaining at "-0:01" (the last per-tick value) — test fails.

### FIX-3 — keyboard-only input pause (test-gap MED #2)

- No code change; existing `scrubber.addEventListener('input', …)` already calls `scrubFromPause()` unconditionally.
- Tests added: `input event with no preceding mousedown still pauses a playing engine (Arrow-key path)`.
- Mutation hypothesis: moving `scrubFromPause()` into a `if (mousedownPending) {…}` branch fails this test.

### FIX-4 — empty-stream max=0 guard (test-gap MED #3)

- Belt-and-braces — `Math.max(0, …)` already at mount; FIX-1's `updateScrubber` guard now defends the dynamic path too.
- Tests added: `empty stream renders scrubber with max="0" and does not crash on update`.
- Mutation hypothesis: removing `Math.max(0, p.total - 1)` from updateScrubber leaves max="-1" after engine transitions to `done` on empty start.

### FIX-5 — styles.ts auto-hide comment (architect MED #2)

- `styles.ts:272-280` — single-block comment ahead of `.scrubber-area` warning #52 polish away from `display:none`.
- **Coordinator-brief deviation:** brief showed the example using backticks (`` `visibility:hidden` ``) in markdown style. The styles.ts file is a TS backtick template literal — backticks inside the CSS comment terminate the template string and break tsc + lint + prettier. Rewrote without code-span backticks (`visibility:hidden` instead of `` `visibility:hidden` ``). Same intent, valid syntax. Per `[[coordinator-briefs-are-auditable]]` flagging the deviation here rather than silently rewriting.

### FIX-6 — live-region storm suppression (a11y MED #1)

- `overlay.ts:597-617` — `scrubInProgress` + `scrubAnnouncementFired` flags + `SCRUB_DEBOUNCE_MS = 250` constant.
- `overlay.ts:142-145` — outer-closure `scrubDebounceTimer` for unmount cleanup (mirrors `resumeToastTimer` pattern).
- `overlay.ts:719,733` — `if (!scrubInProgress) ariaLive.textContent = ev.word` (word branch) and same guard on chunk branch.
- `overlay.ts:1014-1029` — `beginScrubSession()` writes announcement first, flips flag, resets debounce on every event.
- `overlay.ts:1090-1095` — clearTimeout in unmount.
- Tests added (3): rapid-scrub-storm suppression; post-debounce emission unblocked; back-compat normal playback still updates aria-live.
- Mutation hypothesis: removing `if (!scrubInProgress)` from word branch fails test 1 (aria-live overwritten with "fine."). Removing the debounce reset in `beginScrubSession` fails test 2 (the flag stays true forever, blocking the post-debounce update).

### FIX-7 — pause-on-scrub polite announcement (a11y MED #2)

- `constants.ts:144-153` — `SCRUB_PAUSED_ANNOUNCEMENT: 'Paused. Scrubbing reading position.'`.
- Same `beginScrubSession()` helper handles the one-shot — announcement written BEFORE `scrubInProgress` flips true so the suppression doesn't block our own write.
- Tests added (3): first-of-session fires; no re-fire within session; new session after debounce re-fires.
- Mutation hypothesis: removing the `scrubAnnouncementFired` latch fails test 2 (re-fires on every input). Removing the latch reset in the debounce-timer callback fails test 3 (never re-fires after first session).
- **Design choice (per brief — decision point):** reused the existing `.aria-live` node rather than adding a dedicated one. FIX-6's suppression flag guarantees no per-emission word can race ahead of our announcement, so a separate node would be redundant DOM.

## Test delta

1219 → 1230 = **+11 vitest** (24 total in scrubber.test.ts vs 13 prior). 0 regressions across 80 test files.

## Self-pushback / deviations from brief

1. **FIX-5 backtick deviation** documented above (CSS comment inside TS template literal).
2. **Architect HIGH framing** — confirmed `setChunkSize` axis claim is incorrect via direct engine source read; still added the resync as defensive depth + added a pinning test for the mode-invariance promise. Documented in fix comment per brief.
3. No other deviations — followed the brief's per-FIX implementation map.

## CI output tails (from worktree)

```
$ npx tsc --noEmit
(clean, no output)

$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
(clean)

$ npm run format:check
Checking formatting...

$ npm test
 Test Files  80 passed (80)
      Tests  1230 passed (1230)
   Start at  09:37:33
   Duration  3.32s

$ npm run build
✓ built in 224ms

$ npm run build:e2e-bundle
dist-e2e/core-overlay-bundle.js  39.86 kB │ gzip: 13.29 kB
✓ built in 62ms

$ npm run build:e2e-ext
dist-e2e-ext/assets/storage-JgBY6-5A.js  72.70 kB │ gzip: 19.62 kB
✓ built in 219ms

$ npm run test:e2e
  28 passed (30.0s) — 11 skipped (pre-existing activeTab gesture-blocked)
  scrubber.spec.ts: 2/2 still green (input + axe)
```

chriscantu:
# Fix-builder report — PR #204 round-2 (tier-1 critic coverage gaps)

2 ITEMs shipped: ITEM-M3 (chunk-mode storm-suppression test) + ITEM-M1 (touch-path announcement test). No production code touched — both are test-only coverage gaps backing existing FIX-6 / FIX-7 round-1 logic. Items M2 + M4 + D1 split to follow-up issue per coordinator brief.

## Per-ITEM summary

### ITEM-M3 — Chunk-mode storm-suppression test

- Added `describe('createOverlay — scrubber chunk-mode storm suppression (#47 round-2 ITEM-M3)')` block with 2 tests after the FIX-7 block:
  1. `rapid scrub events do NOT spam aria-live with each replacement chunk` — mounts overlay with `chunkSize: 2` and the existing STREAM (8 words → 4 chunks), drives `input` events on scrubber across values 2→6, asserts aria-live retains `SCRUB_PAUSED_ANNOUNCEMENT` (not the chunk text "How are" / etc.).
  2. `after debounce window elapses, next chunk emission DOES update aria-live (chunk-mode unblock)` — same setup, scrub once, advance 300ms past debounce, call `engineOf(holder).seekTo(4)` directly (not via scrubber to avoid re-arming session), assert aria-live is NOT empty (chunk replacement event landed).
- **Chunk-emission timing model used:** confirmed via direct read of `rsvp-engine.ts` `seekToChunk()` (lines ~430-490): in chunk mode the PAUSED-state `engine.seekTo(rawIdx)` snaps the raw word index to its containing chunk and emits a replacement `chunk` event (NOT `word`). Driving via the same `input`-on-scrubber path as the word-mode test naturally exercises the chunk branch because the scrubber's input handler calls `engine.seekTo(target)`, which in chunk mode emits a chunk event into the subscribe handler.

**Mutation verified:** removed `if (!scrubInProgress)` from chunk branch (overlay.ts ~line 773). Test fails:
```
expected 'How are' to be 'Paused. Scrubbing reading position.'
```
Restored after verification.

### ITEM-M1 — Touch-path announcement test

- Added `describe('createOverlay — scrubber touch-path announcement (#47 round-2 ITEM-M1)')` block with 1 test.
- Used `new Event('touchstart', { bubbles: true })` matching existing touchstart-pause test pattern (line 237) since jsdom does not constructor-support TouchEvent reliably.
- Test asserts aria-live holds `SCRUB_PAUSED_ANNOUNCEMENT` after `touchstart` AND after subsequent `input` (announcement fires exactly once, not clobbered by per-emission word "you?").

**Mutation verified:** reordered `beginScrubSession` to flip `scrubAnnouncementFired = true` BEFORE the `if (!scrubAnnouncementFired)` check (the actual textContent write is then dead code). Test fails:
```
expected 'Hello.' to be 'Paused. Scrubbing reading position.'
```
The post-touchstart assertion catches it because aria-live still holds the initial `engine.start()` word emission ("Hello."), confirming the announcement was never written. Restored after verification.

## Test delta

1230 → 1233 = **+3 vitest** (3 new in scrubber.test.ts, total 30 in that file). 0 regressions across 80 test files.

## CI tails (from worktree)

```
$ npm run lint
> speedreader-chrome@0.1.0 lint
> eslint . --max-warnings=0
(clean)

$ npm run format:check
> speedreader-chrome@0.1.0 format:check
> prettier --check "src/**/*.{ts,tsx,js,jsx,json,css,md}"
Checking formatting...

$ npx tsc --noEmit
(clean, no output)

$ npm test
 Test Files  80 passed (80)
      Tests  1233 passed (1233)
   Start at  09:51:21
   Duration  3.59s

$ npm run build
✓ built in 229ms

$ npm run test:e2e -- --grep "scrubber"
  ✓  scrubber.spec.ts:29:3 › input event jumps engine to the new index ... (154ms)
  ✓  scrubber.spec.ts:91:3 › axe-core finds no a11y violations ... (251ms)
  2 passed (1.0s)
```

## Self-pushback / deviations from brief

1. **ITEM-M3 unblock-test path deviation:** brief said "After 300ms debounce, fire another chunk emission → assert aria-live DOES update". I used `engineOf(holder).seekTo(4)` directly rather than another `input` event on the scrubber. Reasoning: an `input` event re-runs `beginScrubSession()` which re-sets `scrubInProgress = true` BEFORE the seekTo's chunk emission lands — that would re-arm suppression and the test would (correctly) see no aria-live update, which is the OPPOSITE of what we want to assert. Calling seekTo directly bypasses beginScrubSession so the post-debounce chunk emission flows through the (now-unguarded) chunk branch. This is the only path that actually exercises the assertion the brief intended.
2. **ITEM-M1 mutation framing:** the brief described "regression setting `scrubAnnouncementFired` BEFORE textContent write". The realistic structural mutation that breaks the test is reordering the assignment AND the latch into a dead `if (!latch)` block (after flipping latch first, the check is always false). Verified concretely — see Mutation verified block above.
3. No other deviations.

PR #232, opened 2026-06-11
docs(#196): spec for SW-owned reading-position store (S3)

### LINKED ISSUE #196: spec: move reading-position store into service worker (S3 follow-up to #48)
## Context

PR #195 (#48 reading-position memory) ring surfaced a privacy finding flagged by security-adversary tier-2: content scripts in any extension context (post-user-gesture activation via `activeTab`) can call `chrome.storage.local.get(null)` and enumerate every `position:<canonicalURL>` key, exfiltrating the user's cross-origin reading history to whichever site they activate the reader on.

User scope initially included "full hash + SW key encryption" as the in-PR fix. The fix-builder pushed back with structural reasoning that was correct on review, so S3 was deferred to this follow-up.

## Why encryption alone doesn't fix it

Both candidate key-storage paths defeat the threat model:

- **`chrome.storage.session` for the AES-GCM key** — SW eviction (~30 s idle in MV3) wipes the key, making encrypted records unreadable. Counterproductive for a persistence feature whose value proposition is bridging sessions. User would see "history vanished" every SW eviction.
- **`chrome.storage.local` for the AES-GCM key** — content scripts read the key alongside the encrypted blobs. SHA-256 URL hashing is a public algorithm with no obstruction value (a hostile CS can compute the same hashes the SW does). Net security improvement vs plaintext: zero.

## The structural fix

Move the reading-position store into the service worker. Content scripts become thin clients that communicate via `chrome.runtime.sendMessage`:

```
SW (data owner)
  ├─ owns chrome.storage.local 'position:*' keyspace
  ├─ message handlers: position/read, position/write, position/clear, position/clear-all, position/list
  └─ no key material, no crypto needed — CS simply has no read path

CS (thin client)
  ├─ ChromeMessagePositionAdapter implements StorageAdapter via sendMessage
  ├─ debounced 1s writes still apply
  └─ pagehide / visibilitychange flush still apply (fire-and-forget sendMessage)
```

Encryption is theater on top of this move; without this move, encryption doesn't help.

## Design questions for the spec PR

1. **SW cold-start race window.** First `position/read` after SW eviction has cold-start latency (~50-200 ms). Does the overlay block the first word render on this, or render at index 0 and patch when the read resolves? The latter avoids latency but flashes word 0 briefly. The former blocks the most-visible interaction. Trade-off worth documenting.

2. **Message ordering during overlay mount.** Read fires on mount; the first write fires ~1 s later (debounce). If reads/writes interleave across overlays mounted in two tabs near-simultaneously, the SW handler MUST serialize writes per-URL (probably reusing the same promise-queue pattern from #48). Spec needs to call this out.

3. **Perf cost at the 1 Hz write rate.** Every write becomes a CS→SW round-trip + IPC. Sustainable at 1 Hz; unsustainable if a future change drops the debounce. Document the constraint.

4. **List() perf for #49 popup.** Returning N decrypted (or now just-read) entries on every popup open is fine for N=100. If LRU cap ever bumps to 1000+, paginate.

5. **Migration from #48's CS-owned `position:*` keys.** The existing keys land before this PR. SW handler reads them on first message and rewrites under the same key shape (no schema change, only ownership change). Document the one-time migration window.

6. **Privacy semantics that survive.** The SW still stores plaintext URLs in `chrome.storage.local`. A user who has the extension force-uninstalled by Chrome (extension corruption, manual storage inspection via DevTools on chrome-extension://) can still see the data. The threat closed by this work is specifically *cross-origin enumeration from a CS context the user opted into via activeTab*, not at-rest disk privacy.

## Acceptance

- Spec PR proposing the SW-owned message API + the six design-question resolutions above.
- Spec merges before impl dispatches.
- Impl PR: SW handler module + message-based `StorageAdapter` impl + integration tests covering round-trip + cold-start + cross-tab serialization + migration from CS-owned keys.

## Related

- PR #195 (issue #48) — surgical hardening (S1+S2+A1+A2+A3+NITs) landed; S3 carved out to here.
- Issue #49 (reading history) — popup consumer of `store.list()`. Should land BEFORE this S3 work so the consumer surface is concrete, then S3 migrates both the store AND the popup binding in one structurally-coherent change.

## Labels

`area:settings`, `area:privacy`, `scope:chrome-port`

chriscantu:
## Summary

Spec for #196 — close the cross-origin reading-history enumeration threat PR #195's ring surfaced (S3, from #48). **Backend pivoted 2026-06-15** from extension-origin IndexedDB to **`chrome.storage.local` + `setAccessLevel({ accessLevel: 'TRUSTED_CONTEXTS' })`**.

**Why the pivot.** The reading positions already live in `chrome.storage.local`. Calling `chrome.storage.local.setAccessLevel('TRUSTED_CONTEXTS')` from the SW removes the `local` area from the content-script API surface entirely — closing the exact `get(null)` enumeration threat **without moving any data**. This dominates the IDB direction on every axis: durable (local survives restart natively), CS-isolated (verified), and **zero data migration** — it deletes the entire crash-safe `local`→IDB purge regime (the most error-prone, irreversible part of the IDB design: a purge of live user reading-history).

Key design points (the message/security layer is unchanged from the IDB draft — it sits above the storage seam):
- 6 new one-shot RPCs split across the existing `CS_ONLY`/`POPUP_ONLY` provenance sets; no content-script message type carries a `url` (the security invariant, made syntactic).
- CS reads bind to `sender.url` (not `sender.tab.url`) — populated without the `tabs` permission, pinned to the top frame by the existing `frameId === 0` gate.
- The SW re-asserts `setAccessLevel` at module top-level every wake. The re-assertion window is closed for newly-injected CS by the no-`content_scripts` lazy-injection model, and for long-lived CS during SW eviction by check 6 (empirically).
- Centralizing the store in one SW instance also closes the #48 cross-tab LRU race as a bonus.

Status: **Proposed** → Accepted on merge. Stress-tested by extension-architect + decision-challenger; all findings addressed (options-page second-writer rebind, fail-closed fallback + `minimum_chrome_version` pin as impl preconditions, area-global `local` foreclosure doc, CDP-stop vs natural-eviction caveat).

Closes #196 (spec deliverable).

## Test Plan

Spec + reproducer-harness change. Reproducer (`experiments/idb-isolation-check/`) results recorded in the spec; gates the Accepted flip.
- [x] `prettier --check` (format:check) — clean
- [x] `eslint .` — clean (exit 0)
- [x] `vitest run` — 1377 passed
- [x] `vite build` — clean
- [x] Empirical reproducer `npm run test:idb` — **13/13 green** (Chrome 148). Check 5 (CS blocked from `local` after `setAccessLevel`, SW alive) + check 6 (restriction persists while SW stopped) + check 2 (`sender.url` binding) all PASS, each with a non-vacuity red-check. Merge gate CLEARED.
- [x] Architect + decision-challenger drafting review — findings incorporated


chriscantu:
## Ring + architect review — complete

Ran the full antagonistic ring (forced LARGE tier; doc-only diff would auto-rate trivial, but the content is a security/privacy/MV3-boundary design) + arbiter + extension-architect on the original draft. **Verdict: HOLD-with-fixes → all findings applied in 257a8d3. Core architecture (IDB direction, sender.url binding, threat model) was affirmed by every reviewer.**

### Convergence-ranked (4-critic ring + arbiter)
- **HIGH** forged-URL regression test was a tautology risk → mandated assertion shape (response == sender record AND != forged record AND forged key untouched), merge-blocking.
- **HIGH** migration purge-sweep underspecified (could omit + still pass) → marker-independent sweep via `storage.local.get(null)` prefix filter, normative.
- **HIGH** cold-start latency underestimated (migration-wake 300-500ms) → blocking default gated on a reproducer measurement; flip to render-then-patch / async migration if it regresses.
- **HIGH** cross-tab race test can't distinguish fixed-from-broken in sinon-chrome → proved-by-architecture + multi-tab E2E.
- **MED** storage.session rejection rested on an unsourced claim → reproducer-gated (check 3).
- **MED** popup-direct-IDB ambiguity → RPC normative (direct write reintroduces the multi-writer LRU race).
- **MED** type-confusion (gate is blocklist) + sender.url-null poisoning → fail-closed gate default + null→Result.err+no-write.

### extension-architect (structural, zero overlap with the ring)
- **HIGH** migration trigger "first activation" isn't a real lifecycle event → runs every wake, not onInstalled-only.
- **MED** onMessage async handlers must use the IIFE + sync-return-true shape; route() stays void (no test catches a dropped sendResponse).
- **MED** migrate.ts must receive the shared IDB connection, not open its own.
- **LOW-MED** handlers must await the shared open-connection promise even warm.
- **LOW** gate fail-closed default as a code invariant.

### Design tensions adjudicated
- Popup RPC vs direct IDB → **RPC** (single-writer correctness beats the saved IPC).
- Cross-tab test → **proved-by-architecture**, E2E only.
- Partial-purge coverage → **kept** (security makes the sweep load-bearing); unit asserts logic given a simulated abort, E2E is the real-death source of truth.

Status stays **Proposed**. The empirical reproducer (now 3 checks: IDB-isolation, sender.url, session-restart) runs inside this PR and gates merge + the Accepted flip. No impl dispatches until it's green.

PR #235, opened 2026-06-16
feat(#196): SW-owned reading-position store (setAccessLevel gate, sender-bound RPCs)

chriscantu:
## Summary

Implements #196 — the SW-owned reading-position store. Closes the cross-origin
reading-history enumeration threat (#195/S3) flagged by PR #195's ring: a content
script could call `chrome.storage.local.get(null)` and exfiltrate every
`position:<url>` key. The SW now calls
`chrome.storage.local.setAccessLevel({accessLevel:"TRUSTED_CONTEXTS"})` at module
top-level on every wake, removing `local` from the CS API surface. CS/popup/options
become thin clients reaching the SW-owned store via sender-bound `sendMessage` RPCs.

**No data migration** — positions stay in `chrome.storage.local`; the existing
`chrome-position-store.ts` adapter is reused unchanged, now SW-side.

Spec (merged, Accepted): `docs/superpowers/specs/2026-06-11-position-store-service-worker.md`

### Key decisions
- **`minimum_chrome_version` = 140**, not the spec's untested "≥119 expected". MDN
  browser-compat-data is explicit: `setAccessLevel` on the `local` area is
  "Supported by all storage areas from Chrome 140" (session got it at 96/102).
  This is the impl-PR floor confirmation the spec delegated. Fail-closed
  feature-detection means security holds at any version regardless of the pin.
- **Fail-closed**, never permissive: if `setAccessLevel` is absent, the store
  refuses to persist (no resume) rather than write `position:*` to un-gated `local`.
- **Single-writer rule**: no trusted context (popup OR options) remains a direct
  `chrome.storage.local` writer — all mutations route through the SW RPC.
- **Fail-closed gate default**: a message type in neither provenance set is
  rejected (`invalid-payload`), not forwarded to the router by sender shape alone.

## Test Plan

- [x] `npm run lint` — clean (0 warnings, `--max-warnings=0`)
- [x] `npm test` — 1450 passed (incl. forged-URL mandatory-shape, null/opaque
      reject-no-write, all-6-types provenance rejection, fail-closed zero-writes,
      access-gate ordering/feature-detect, CR2 resume over RPC)
- [x] `npm run build` — tsc + vite clean; `minimum_chrome_version: "140"` in dist manifest
- [x] E2E isolation assertion on the REAL loaded extension (Chrome 148):
      `RUN_POSITION_ISOLATION=1 npx playwright test tests/e2e/position-isolation.spec.ts`
      → 1 passed. CS keyed-get + `get(null)` enumeration BOTH fail; trusted SW read
      survives (wordIndex 7); CS saw zero position keys. Non-vacuous (seeded key
      would surface if the gate were absent).
- [ ] **Manual, once-per-Chrome-major (non-blocking, deferred):** confirm the
      eviction-persistence (check 6) under NATURAL idle-eviction (~30-40s, no CDP)
      on the min-supported Chrome (140). Cannot be automated in CI; flagged per
      spec §Empirical Precondition check-3 faithfulness caveat.

## Notes
- Reproducer (`experiments/idb-isolation-check/`) already proved checks 5/2/6
  (13/13). This PR adds the runtime proof inside the production extension.
- E2E specs are opt-in (`RUN_POSITION_ISOLATION=1` + `npm run build:e2e-ext`),
  matching the repo convention for extension-loading specs.

### DIFF
--- a/src/chrome/__tests__/manifest.test.ts
+++ b/src/chrome/__tests__/manifest.test.ts
@@ -10,6 +10,19 @@ import manifest from '../manifest';
  * binary lands separately (see fonts/README.md); the manifest contract
  * MUST be in place first so #27/#28 don't have to touch this surface.
  */
+/**
+ * #196 — `minimum_chrome_version` is pinned to the `local.setAccessLevel`
+ * floor (Chrome 140 per MDN browser-compat-data). This is a security-load-
+ * bearing merge precondition: below the floor the gate API is absent and the
+ * cross-origin reading-history enumeration threat would reopen. A regression
+ * that drops or downgrades this string must fail CI, not ship green.
+ */
+describe('manifest — minimum_chrome_version floor (#196)', () => {
+  it('pins minimum_chrome_version to the local.setAccessLevel floor (140)', () => {
+    expect(manifest.minimum_chrome_version).toBe('140');
+  });
+});
+
 describe('manifest — web_accessible_resources for fonts (#10)', () => {
   const wars = manifest.web_accessible_resources ?? [];
   const fontEntry = wars.find((entry) =>
--- a/src/chrome/background/__tests__/route.test.ts
+++ b/src/chrome/background/__tests__/route.test.ts
@@ -213,6 +213,39 @@ describe('route — activate-reader handler', () => {
     expect(deps.dispatchActivation).not.toHaveBeenCalled();
   });
 
+  // #196 — route delegates the six position types to handlePosition before
+  // the activate-reader branch. (In this unit env the default position store
+  // is the fail-closed null store, since no chrome.storage.local is present.)
+  it('dispatches popup position/list (returns ok with the default store)', async () => {
+    const { promise, sendResponse } = waitForResponse();
+    route(
+      { type: 'position/list' },
+      popupSender(),
+      sendResponse,
+      makeDeps({ ok: true, data: undefined }),
+    );
+    const resp = await promise;
+    expect(resp.ok).toBe(true);
+  });
+
+  it('dispatches CS position/get bound to sender.url (returns ok)', async () => {
+    const csSender: chrome.runtime.MessageSender = {
+      id: OWN_ID,
+      url: 'https://example.com/article',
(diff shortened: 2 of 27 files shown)


chriscantu:
## Antagonistic ring — tier:hard (4 critics + arbiter-inline + re-review, opus)

Per `feedback_sonnet_false_mutation_evidence` (always-ring behavior-touching security PRs). Convergence-ranked.

| Finding | Dim | Sev | Verdict |
|---|---|---|---|
| Gate fail-OPEN on `setAccessLevel` runtime rejection (presence-detect, not resolution) | security | HIGH | **FIXED** → resolution-gated, live per-op signal; reject+sync-throw fail closed |
| Options page (opens in a tab) mis-classified as CS → `position/clear-all` rejected at gate | security | MED | **FIXED** → classify trusted by `chrome-extension://` origin |
| `minimum_chrome_version` 140 had no regression test | test-gap | MED | **FIXED** → manifest test |
| access-gate ordering test was a tautology | test-gap | MED | **FIXED** → resolution-ordering test |
| client `no-response` (SW-asleep) branch untested | test-gap | LOW | **FIXED** → coverage added |
| `position/delete` canonicalization delegation undocumented | security | LOW | **FIXED** → defense-in-depth comment |
| CR2 `wordIndex` range-guard boundary untested | test-gap | LOW | **FIXED** → boundary test |
| Write path 4-5 storage ops / redundant get under queue lock | perf | LOW | **WONTFIX** — pre-existing #48 adapter code in a spec-`UNCHANGED` file, spec-accepted §197; follow-up only |

**Scope critic:** clean — every line traces to #196 §Code Layout; the two UNCHANGED files (chrome-position-store.ts, core reading-position.ts) are byte-for-byte untouched; no migration code.

**Security re-review verdict:** findings #1/#2/#3 **CLOSED**, no new security hole; web CS cannot forge `chrome-extension://` origin (Chrome-populated `sender.url`).

Final: 1467 tests pass, lint/format/build green, E2E isolation re-verified on Chrome 148.

